PR #55, opened 2026-05-13
feat: add POST /execute/fetch endpoint for browser-side fetch proxy

gabriel-adopt:
## Summary

- Adds `POST /execute/fetch` to the worker health server (port 8091) — runs `fetch()` inside the authenticated browser via `page.evaluate()` with `credentials: 'include'`, returning status/headers/body
- Adds K8s Service `{pod_name}-worker` on port 8091 so the API can route to the worker via DNS. Updates NetworkPolicy to allow API ingress on 8091
- Adds `ExecuteModule` to the NestJS API — resolves profile → session → pod, forwards to worker, with per-profile rate limiting (60 req/min via Redis) and full error mapping (409/429/502/504)
- Shared types (`ExecuteFetchRequest`/`ExecuteFetchResponse`, `EXECUTE_LIMITS`) in `packages/shared`

## Architecture

```
NoUI MCP Server → POST /execute/fetch → Tabby API (auth + routing) → worker pod (page.evaluate(fetch(...))) → response
```

No WebSocket. No CDP port exposure. No `Runtime.evaluate` from outside the pod.

## Plan reference

Implements Phases 1-2 of the tabby-execute-endpoint-plan. João's integration notes resolve all open questions (port reuse, binary responses deferred to v2, no mutex v1, CORS as documented limitation).

## Test plan

- [x] `pnpm run build` — all 7 packages compile
- [x] `pnpm run test` — 689 tests pass (4 suites: API 506, controller 54, worker 48, shared 81)
- [x] `pnpm run lint` — clean
- [x] Integration test against staging (manual — needs deployed worker with execute handler)
- [x] Security validation: scheme rejection, auth enforcement, no cookie/localStorage leakage


joaomoraski-adoptai (commented):
## Review: `feat/execute-endpoint`

Overall the feature is well-structured — the API-side validation, rate limiting, and browser lock are solid foundations. A few security, correctness, and architecture issues need to be addressed before this ships to production.

---

### 🔴 [SECURITY] Authentication on worker execute endpoints — `apps/worker/src/health-server.ts`

`registerExecuteHandler` and `registerBrowserHandler` are called unconditionally in `setPage`, registering unauthenticated routes on port 8091. The worker trusts any request that reaches that port, relying solely on the NetworkPolicy. That's not sufficient — if `execute_enabled` is ever misconfigured or the NetworkPolicy has a gap, any pod in the cluster can issue arbitrary browser commands.

**Recommendation:** Use Tabby's existing token exchange flow for authentication. The platform already exchanges its JWT for a Tabby-scoped JWT (NoUI/Agent Canvas need to use this path) via `POST /auth/token-exchange` (see `apps/api/src/modules/auth/token-exchange.service.ts:50` and `apps/api/src/modules/auth/auth.controller.ts:406`). The validation of the JWT can be in the Tabby-scoped one, so other types of auths in tabby, like the Admin Login can use this endpoint too

Take a look at how `adoptwebui` calls `tabby_resolution_service.py` → `/auth/token-exchange` → `/credentials/request` to see the full flow in production.

Avoid relying solely on agent `client_id`/`client_secret` for this — that path will be deprecated in the future. Token exchange is the canonical auth mechanism going forward.

For local dev and testing, using the Tabby admin user credentials (`POST /auth/login`) is fine. But in production, the execute endpoints must validate a proper Tabby JWT.

Also important: local NoUI connecting to a cloud Tabby instance should not be supported — it introduces cost tracking and security concerns that aren't worth solving at this stage.

---

### 🔴 [SECURITY/ARCHITECTURE] Execute feature should be opt-in per Application — `apps/controller/src/reconcile.service.ts`

`createWorkerService` (line ~701) is called unconditionally for every session, regardless of whether the application needs execute capabilities. This means every VNC and CDP session gets the worker Service + the execute routes registered, even when they'll never be used.

This should be gated behind a flag on the Application config (e.g., `execute_enabled: boolean`):
- **Controller**: only create the worker Service when `execute_enabled = true`
- **Worker**: only register `/execute/*` routes when the flag is active
- **NetworkPolicy**: only open port 8091 when needed

Without this, every session exposes the execute surface — including sessions for other tenants/clients. Even with NetworkPolicy, defense in depth requires minimizing the attack surface to only the sessions that actually need it.

---

### 🔴 [BUG] `executeBrowser` doesn't check `workerResponse.ok` — `apps/api/src/modules/execute/execute.service.ts`

In `executeBrowser` (~line 439), after calling `fetch(workerUrl, ...)`, the code does:

```ts
const result = await workerResponse.json() as ExecuteBrowserResponse;
return result;
```

Unlike `executeFetch` which has `if (!workerResponse.ok)` guard, this method calls `workerResponse.json()` unconditionally. If the worker returns HTTP 4xx/5xx, the API returns 200 with `success: false` — the caller can't distinguish a network failure from a command failure. Add the same `!workerResponse.ok` guard here.

---

### 🟡 [BUG] Rate limit charged before URL validation — `apps/api/src/modules/execute/execute.service.ts`

In `executeFetch`, `enforceRateLimit(profileId)` is called at line ~306, but `validateRequest(request)` is called later at line ~321. Malformed requests (invalid URL, bad scheme, oversized body) burn a rate limit slot before being rejected. Move `validateRequest` before the rate limit check so invalid requests don't penalize the caller's quota.

---

### 🟡 [BUG] HAR `requestMap` key collision — `apps/worker/src/execute-browser-handler.ts`

In the `har_start` case, the request key is built as:
```ts
requestMap.set(`${request.method()}:${url}:${startTime}`, ...)
```

But the response lookup does:
```ts
const key = [...requestMap.keys()].reverse().find(k => k.startsWith(`${request.method()}:${url}:`));
```

The timestamp suffix is ignored on lookup. Two concurrent requests to the same URL within the same millisecond will collide — the second response handler will find/delete the same entry, and one HAR entry gets silently dropped.

Use a UUID or monotonic counter instead of `Date.now()` for the request key, and carry it via Playwright's `request` object (e.g., attach it to a WeakMap keyed by the request object).

---

### 🟡 Caller-supplied headers forwarded without sanitization — `apps/worker/src/execute-handler.ts`

In `page.evaluate(...)`, the caller's `headers` object is forwarded directly:
```ts
headers: h,
```
Combined with `credentials: 'include'`, a caller can override `Cookie`, `Authorization`, or `X-CSRF-Token` headers that the browser session already has.

This is partially mitigated by the API-layer ownership check (callers can only target sessions they own), but within that scope there's no restriction on what headers can be injected. Consider documenting this as intentional behavior or adding a blocklist for sensitive headers.

---

### 🟡 Module-level `activeHar` singleton — `apps/worker/src/execute-browser-handler.ts`

```ts
let activeHar: HarCapture | null = null;
```

`activeHar` is a module-level singleton. Currently fine since there's one session per worker process, but if `setPage` were ever called again (e.g., after page navigation/recreation), `detachHarListeners` would reference the old page. Consider scoping this to the handler instance or the page object.

---

### 🟡 Missing tests — `apps/worker/src/execute-handler.ts` + `execute-browser-handler.ts`

Both files add ~430 lines of new logic with zero test coverage. Given that these are new endpoints that execute arbitrary fetch/browser commands, unit tests are needed — at minimum covering:
- URL validation and scheme rejection
- Rate limit enforcement
- HAR entry capture and concurrent request handling
- Timeout behavior
- Error response formatting

---

### 🟡 `HAR_BODY_LIMIT_BYTES` unused — `packages/shared/src/execute-types.ts`

```ts
HAR_BODY_LIMIT_BYTES: 10_485_760, // 10MB
```

This constant is defined but never referenced. The handler uses `MAX_RESPONSE_BODY_BYTES` (5MB) for HAR body truncation instead. Remove the dead constant or wire it up.

---

### 🟡 [SCALING] INCR/EXPIRE race in rate limiter — `apps/api/src/modules/execute/execute.service.ts`

Both `enforceRateLimit` and `enforceBrowserRateLimit` use the pattern:
```ts
const count = await this.redis.incr(key);
if (count === 1) {
  await this.redis.expire(key, RATE_LIMIT_WINDOW_SECONDS);
}
```

If the process crashes between `INCR` and `EXPIRE`, the key has no TTL and the rate limit gets permanently stuck. Consider using a Lua script or `SET key value EX ttl NX` pattern for atomic rate limiting.

---

### ℹ️ `reconcileRuntimeDrift` orphan cleanup gap — `apps/controller/src/reconcile.service.ts`

Unlabeled orphan pods are deleted (line ~388) without calling `deleteWorkerService` — the `{podName}-worker` Service would be orphaned. This is a pre-existing gap (same pattern exists for noVNC/CDP Services in the unlabeled path) but now affects an additional resource type. Consider adding `deleteWorkerService` to the unlabeled orphan cleanup path.

---

### ℹ️ Breaking change — existing sessions after deploy

Sessions created before this deploy won't have the `{podName}-worker` Service (see `buildWorkerServiceName` in `pod-manager.service.ts`). Execute requests against those sessions will fail with DNS resolution errors. This is expected (sessions recycle naturally), but worth noting in release notes so operators know the feature only works for sessions created after the deploy.

joaomoraski-adoptai (commented):
## Follow-up review

Tested the Tabby platform integration (VNC + CDP sessions, credential resolution, HITL flow) — **everything works normally**. The `execute_enabled=false` default correctly prevents the worker Service and execute routes from being created on existing sessions. After addressing the points below, this is ready to merge.

---

### 1. Worker auth should validate session identity (hardening)

`apps/worker/src/execute-auth.ts` — the middleware calls `jwt.verify()` but doesn't check any claims. The token already carries `tenant_id` and `profile_id` (see `execute.service.ts:336`), but the worker doesn't verify these match its own `TENANT_ID` / `SESSION_ID` env vars.

In the normal flow this isn't exploitable (the API resolves the correct worker via DNS), but for defense in depth the worker should confirm "this token is meant for me":

```typescript
const payload = jwt.verify(token, signingKey, { algorithms: ['HS256'] }) as any;
if (payload.tenant_id !== process.env.TENANT_ID) {
  res.status(403).json({ error: 'Token tenant mismatch' });
  return;
}
```

This is a 3-line change.

---

### 2. Tests should include integration coverage

The new test files (`execute-handler.spec.ts`, `execute-browser-handler.spec.ts`, `execute-auth.spec.ts`) are pure unit tests — they validate parsing logic and constants but don't instantiate Express, don't mock Playwright, and don't test the actual handler functions.

The existing Tabby test pattern (see `credentials.spec.ts` with 56 tests covering service methods, schema validation, profile/session resolution, and volatility classification) uses integration-style tests that exercise the real service layer. The execute handlers should follow the same pattern — at minimum covering the auth middleware rejecting bad tokens, the rate limiter returning 429, and the handler returning proper error shapes.

---

### 3. Migration not registered in `data-source.ts`

`apps/api/src/migrations/1708300000022-AddExecuteEnabled.ts` exists but is not imported or listed in `apps/api/src/data-source.ts`. TypeORM won't run it on startup. Add the import and include it in the `migrations` array.

---

Once these are addressed, good to merge. Nice work on the `execute_enabled` gating and the Lua rate limiter.

joaomoraski-adoptai (approved):
lgtm

PR #105, opened 2026-06-16
feat: VNC session recording (login + workflow) with egress allowlist closure

gabriel-adopt:
## Summary

Adds **VNC session recording** to Tabby — an agent generates a viewer URL, a human drives a Tabby-hosted browser, and the worker captures HAR + DOM interactions server-side, draining a bundle NoUI compiles into a login profile or an MCP. This is the Tabby half of replacing the NoUI Chrome extension. It also closes the **egress allowlist** loop that the recording feature depends on.

Base is `tabby-noui` (where this was cut from); 13 commits.

## What's in it

**Recording engine (worker)**
- Ambient HAR + DOM interaction + URL capture during a human VNC session; in-pod credential redaction; `/execute/*` blocked in recording mode.
- DOM interactions are emitted over a **network beacon** (sentinel `fetch()` read via `page.on('request')`) — the only CDP channel the CloakBrowser stealth build forwards (exposeBinding + console are stripped). Injected via `addInitScript` + per-navigation `page.evaluate`.
- Cookies captured at drain → **authenticated session reuse**: a workflow recording provisioned `source_session_id` (a prior login recording) is seeded via `context.addCookies()`, so the human starts already signed in — no stored credentials.

**Recording API + provisioning**
- `POST /recording/sessions` (agent bearer) provisions a recording-shell app + session and returns a `?mode=recording` VNC URL.
- `POST /vnc/:id/recording-stop` drains → AES-256-GCM encrypted persist to MinIO; `GET /recording/sessions/:id/bundle` retrieval; single-use stream-token refresh.
- Controller provisions the worker-health Service (drain endpoint) for recording sessions, not just execute apps.

**Egress allowlist** (delivers the [egress proposal](../plans/tabby/egress-allowlist-fix-proposal.md) Layer 1 + scoped Layer 2 — see the attached handoff `vnc-recording-egress-handoff.md`)
- `.adopt.ai` hardcoded infra allowlist in the egress proxy.
- `extra_egress_allowlist` on `applications` + `app_templates` (migration 026), composed into the per-session allowlist by the controller; cloned template→app.
- Recording sessions run **allow-all** egress (a `*` sentinel); the HAR-derived domains then populate `extra_egress_allowlist` for enforce-mode replay.

**VNC perf** — worker CPU 1.5→4, x11vnc `-threads`, 1440×900, noVNC compression (multi-minute first paint → seconds).

## Why

Replaces the extension recording flow; makes authenticated recording + replay work without per-vendor `kubectl set env` (the egress proposal's core problem) and without storing credentials (captured-session reuse).

## Risk

Medium. New worker recording path + new API/controller surface; existing login/execute paths untouched. Egress changes are additive (new field + always-allow + per-session allow-all for recording only). Migration 026 is additive (jsonb default `[]`).

## Validation

- Full suite green: api / worker / controller, incl. new recording-runner (beacon channel) + pod-manager (egress PUT body) tests.
- E2E on Kind: login recording (interactions captured, password redacted) → compiled login profile (8-step DSL, 28-domain `extra_egress_allowlist` from HAR) → authenticated workflow recording (`Seeded 38 cookie(s)`) → MCP with 25 tools incl. the Expedia GraphQL search.

## Egress handoff

`plans/tabby/vnc-recording-egress-handoff.md` (attached) maps this PR to the egress proposal/review for the proposal author: Layer 1 + scoped Layer 2 + NoUI auth-domain wiring shipped; Layer 3 telemetry, `auth_provider`, `egress_mode`/learn mode deferred (learn mode is obviated for the NoUI flow by allow-all recording + HAR-derived allowlist).


joaomoraski-adoptai (commented):
Quick review — looks solid overall. A few notes:

### `extra_egress_allowlist` — controller wiring
The field is stored on apps and templates, and auto-provisioned apps inherit it from the template (via `credentials.service.ts`). However, the controller's `syncEgressAllowlist()` still only passes `app.target_urls` to the egress proxy — `extra_egress_allowlist` isn't merged in yet. This means the field has no runtime effect on the egress proxy right now.

Is this intentional (wiring comes in a follow-up), or should this PR include the controller-side composition (`[...target_urls, ...extra_egress_allowlist]`)?

The controller entity (`apps/controller/src/entities/application.entity.ts`) also doesn't have the `extra_egress_allowlist` column yet, so the controller can't read it even if the sync logic was added.

### RecordingStore — encryption
Nice — AES-256-GCM with the tenant encryption key, same blob format as artifacts. The roundtrip test is good. One minor thing: `drainFromWorker` has a 65s timeout which is generous — is there a case where the worker takes that long to flush?

### Recording provision — cookie seeding
The `source_session_id` flow for workflow recordings inheriting login cookies is clean. Good error message when source has no cookies.

### Base branch
Looks like this was moved to `dev` already ✅. Just noting: `tabby-noui` was synced with `dev` today so it's up to date if you need it later.

joaomoraski-adoptai (commented):
### Suggested fix for egress allowlist composition

If you want `extra_egress_allowlist` to have runtime effect in this PR, here's what's needed:

**1. Controller entity** (`apps/controller/src/entities/application.entity.ts`) — add the column:

```typescript
@Column({ type: 'jsonb', default: () => `'[]'` })
extra_egress_allowlist: string[];
```

**2. Reconcile service** (`apps/controller/src/reconcile.service.ts`) — merge the lists when syncing egress:

```diff
- await this.podManager.syncEgressAllowlist(session.id, app.target_urls);
+ const egressUrls = [...(app.target_urls || []), ...(app.extra_egress_allowlist || [])];
+ await this.podManager.syncEgressAllowlist(session.id, egressUrls);
```

Same change in `createNetworkPolicy` call (~line 308):

```diff
- await this.podManager.createNetworkPolicy(savedSession.id, podName, app.target_urls, streamingMode, app.execute_enabled);
+ const egressUrls = [...(app.target_urls || []), ...(app.extra_egress_allowlist || [])];
+ await this.podManager.createNetworkPolicy(savedSession.id, podName, egressUrls, streamingMode, app.execute_enabled);
```

If this is intentionally deferred to a follow-up, no worries — just wanted to surface it.

joaomoraski:
test

joaomoraski (approved):
test

gabriel-adopt:
Thanks for the review — and the suggested fix; your instinct to merge the two lists is exactly what the PR does. I think both notes were written against dev's pod-manager.service.ts rather than the branch's — the controller wiring is all in commit a0e377e (in this PR's commit list; not in dev). Point by point:

1. extra_egress_allowlist — controller wiring (it's already wired)

Both items from your suggested fix are already in the diff:

- Controller entity — apps/controller/src/entities/application.entity.ts already has it (the build wouldn't compile otherwise, since reconcile reads it):
@Column({ type: 'jsonb', default: () => `'[]'` })
extra_egress_allowlist: string[];
- Reconcile sync — already merged at both call sites:
const { extraAllowlist, allowAll } = this.resolveEgressOptions(app);  // extraAllowlist = app.extra_egress_allowlist
await this.podManager.syncEgressAllowlist(session.id, app.target_urls, extraAllowlist, allowAll);
// …and the same args on createNetworkPolicy

So per-session egress is target_urls + extra_egress_allowlist at create and every reconcile tick — live at runtime. If you're seeing only target_urls in pod-manager.service.ts, that's dev (L255); on the branch it's L262–264.

One deliberate difference from your snippet: I pass extra_egress_allowlist as a separate field to the proxy instead of concatenating it into target_urls. That's load-bearing — the proxy's PUT /allowlist runs target_urlet() (new URL(...)), so a bare suffix like .expedia.com fails toparse and is dropped; extra_allowlist goes through normalizeHost(), which keeps host + leading-dot suffix patterns. NoUI emits exactly those suffixes         (.expedia.com, .trvl-media.com), so [...target_urls, ...extra_egparsed arg would silently lose them. Same union you want, justtype-correct for both URL and suffix inputs — and it also carries the allow_all flag for recording sessions.                                                  

2. drainFromWorker 65s timeout — it's a safety ceiling, not an expected duration. The drain is all in-memory (context.cookies() + stop HAR listeners + in-pod sanitize + assemble) and finishes well under a second even on the 500+-entry / 38-cookie sessions we tested. 65s just avoids a premature abort if a worker is briefly busy or the HAR is unusually large. I'll tighten it to ~30s if you'd prefer.


Net: nothing to change on the wiring — happy to walk through a0e377e live, or I can drop an inline comment on the proxy's PUT /allowlist handler flagging the extractHostFromTarget vs normalizeHost split so the distinction is visible right where it matters.

joaomoraski-adoptai (commented):
You're right on both points — I was reviewing against the truncated diff and missed the controller changes in a0e377e. Apologies for the noise.

The separate `extraAllowlist` param to the proxy makes sense — the URL vs suffix parsing distinction is load-bearing. Thanks for explaining.

LGTM from my side. 🚀

joaomoraski (approved):

PR #119, opened 2026-07-06
feat(execute): attach captured request headers to execute/fetch server-side

gabriel-adopt:
## Problem

Agent-harness `call_web_api` calls to QuickBooks Online GraphQL fail with:

```
502 Browser fetch failed: page.evaluate: TypeError: Failed to fetch
    at https://plugin.intuitcdn.net/web-shell/...js
```

`/execute/fetch` always runs the request as an in-page `page.evaluate(fetch)`, so cookies ride via `credentials:'include'`, but a **client-minted bearer** (e.g. QBO web-shell's silent-refresh `Intuit_APIKey` token, harvested via `request_header_allowlist`) is only attached if the *caller* supplies it. Callers that can't hold that bearer — notably the agent-harness `call_web_api` path, which forwards only the per-user Tabby bearer + LLM headers + cookies — hit QBO's wrapped `fetch` with cookies only, which throws `Failed to fetch`.

The standalone NoUI skill path already solved this (it calls `resolve_auth()` → `/credentials/request` and merges the captured header). The gap is that this attach logic lives *in each caller*, so any caller that forgets it (harness today; MCP previously) reintroduces the bug.

## Change

Centralize captured-credential attachment **in Tabby**, where the credentials already live. New opt-in on `POST /execute/fetch`:

- `attach_captured_credentials` (bool) — when set, the API resolves the profile's captured request headers **for the same session the fetch runs in** and merges them into the outgoing headers server-side. The bearer never has to be supplied by, or exposed to, the caller.
- `refresh_credentials` (bool) — force an immediate re-extraction first (for volatile silent-refresh bearers that rotate faster than `refresh_interval_seconds`).

Details:
- `credentials.service.ts::getCredentialsForSession(profile, session, ...)` reuses the extract-lock / decrypt / `buildCredentialSet` path for an **already-resolved** session — `executeFetch` resolves profile+session once and passes them in, so we never re-resolve a *different* session with a mismatched (session-specific) bearer.
- `execute.service.ts` merges captured headers + CSRF; **caller-supplied headers win** on a case-insensitive name collision (preserves the existing "caller headers forwarded as-is" contract); cookies are excluded (inherited via the session, and browsers forbid a `Cookie` header on `fetch`). Fails soft to caller headers on lookup error, warns when capture is empty.
- Keeps the fetch inside the authenticated browser session (inherits cookies/TLS), so it avoids the anti-fraud risk of a server-side replay.

Fully backward compatible — behavior is unchanged unless `attach_captured_credentials` is set.

## Validation

Verified live end-to-end against the **real QBO sandbox** in a local Kind cluster (manual VNC login → healthy `quickbooks-sandbox` session with the `Intuit_APIKey` bearer captured):

| Call | Result |
|---|---|
| AP GraphQL **without** the flag | `502` — `Browser fetch failed: page.evaluate: TypeError: Failed to fetch @ plugin.intuitcdn.net/web-shell` (reproduces the bug) |
| Same call **with** `attach_captured_credentials: true` | `200` — GraphQL returned **63 real AP transaction rows** |

- Build ✅, `tsc --noEmit` lint ✅, **585/585 API tests** ✅ (incl. 5 new `execute.service.spec.ts` covering inject / caller-wins / opt-out / fail-soft / refresh).

## Companion PR (harness wiring)

The agent-harness consumer is wired up in **adoptai/adoptai-workflows#1414**: `call_web_api`'s dispatch (`_dispatch_call_web_api`) sets `attach_captured_credentials=true` on the `/execute/fetch` call **server-side**. It's always-on, which is safe precisely because this PR fails soft — a profile with no captured headers just gets the caller's headers forwarded unchanged.

Note: the agent/LLM does **not** pass any new parameter, and NoUI-generated recipes need no change — the flag is attached below the agent in the harness worker. This keeps the "no caller can forget the bearer" property that motivated centralizing the attach here rather than in each caller.

Merge order: this PR first (the enabler), then #1414.

### DIFF
--- a/apps/api/src/modules/credentials/credentials.service.ts
+++ b/apps/api/src/modules/credentials/credentials.service.ts
@@ -239,9 +239,13 @@ export class CredentialsService {
 
     // Optionally trigger an immediate re-extraction for volatile bearers (silent-refresh
     // tokens rotate faster than refresh_interval_seconds). Coalesced across concurrent callers.
+    // Only the coalesce leader actually triggers an extraction, so only the leader should
+    // bypass the cache — followers would just do a redundant MinIO round-trip for identical data.
+    let bypassCache = false;
     if (forceRefresh) {
       const coalesceResult = await this.acquireExtractLock(tenantId, profile.profile_id, 'default');
       if (coalesceResult.isLeader) {
+        bypassCache = true;
         await this.signalWorkerExtract(session.id);
         if (waitSeconds > 0) {
           await this.redis.blpop(REDIS_KEYS.extractDone(session.id), waitSeconds);
@@ -249,7 +253,7 @@ export class CredentialsService {
       }
     }
 
-    const bundle = await this.fetchAndDecryptLatestBundle(session, tenantId, consumerId, forceRefresh);
+    const bundle = await this.fetchAndDecryptLatestBundle(session, tenantId, consumerId, bypassCache);
     return this.buildCredentialSet(profile, includeVolatile, bundle?.decrypted);
   }
 
--- a/apps/api/src/modules/credentials/credentials.service.ts
+++ b/apps/api/src/modules/credentials/credentials.service.ts
@@ -239,18 +239,22 @@ export class CredentialsService {
 
     // Optionally trigger an immediate re-extraction for volatile bearers (silent-refresh
     // tokens rotate faster than refresh_interval_seconds). Coalesced across concurrent callers.
-    // Only the coalesce leader actually triggers an extraction, so only the leader should
-    // bypass the cache — followers would just do a redundant MinIO round-trip for identical data.
     let bypassCache = false;
     if (forceRefresh) {
       const coalesceResult = await this.acquireExtractLock(tenantId, profile.profile_id, 'default');
       if (coalesceResult.isLeader) {
+        // Only the leader actually triggers the extraction; it reads past the cache.
         bypassCache = true;
         await this.signalWorkerExtract(session.id);
-        if (waitSeconds > 0) {
-          await this.redis.blpop(REDIS_KEYS.extractDone(session.id), waitSeconds);
-        }
       }
+      if (waitSeconds > 0) {
+        // Caller wants fresh data: every caller (leader OR a concurrent follower) waits for
+        // the extraction to land and then bypasses the cache, so nobody gets a stale bundle.
+        await this.redis.blpop(REDIS_KEYS.extractDone(session.id), waitSeconds);
+        bypassCache = true;
+      }
+      // waitSeconds == 0 is fire-and-forget: a follower keeps bypassCache=false and reuses
+      // the cache rather than doing a redundant MinIO round-trip for identical data.
     }
 
     const bundle = await this.fetchAndDecryptLatestBundle(session, tenantId, consumerId, bypassCache);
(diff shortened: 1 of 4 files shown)


joaomoraski-adoptai (commented):
3 minor observations 

**1. `execute.service.ts` — `validateRequest` runs before `mergeCapturedHeaders`**

`validateRequest` enforces `MAX_HEADER_COUNT` on the caller's headers (line 70), but `mergeCapturedHeaders` runs after and can add captured headers beyond that limit. Low severity since captured header count is template-controlled (not untrusted input), but worth knowing if you want the limit to be authoritative. A post-merge count check would close the gap, or just leave as-is if template authors are trusted.

**2. `credentials.service.ts` — `attach_captured_credentials` silently no-ops without `credential_types.headers` on the profile**

`buildCredentialSet` (line ~702) only emits headers when `credential_types.headers` is `'ALL'` or a specific array. If a profile has `request_header_allowlist` configured (captures bearers at the worker level) but `credential_types.headers` is missing, `getCredentialsForSession` returns `headers: []` and the bearer is never injected. Feature looks like it works (200 OK) but the downstream request fails with 401. Worth a note in the Swagger description or a louder warning that names the missing field — easy misconfig for someone setting up a new profile.

**3. `credentials.service.ts:252` — Non-leader passes `bypassCache=true` unnecessarily**

When `forceRefresh=true` but this caller lost the coalesce lock (non-leader), `fetchAndDecryptLatestBundle` is still called with `forceRefresh` (4th arg = bypassCache). The non-leader didn't trigger an extraction, so MinIO holds the same data as the cache — the bypass just adds an unnecessary MinIO round-trip. Passing `coalesceResult.isLeader` instead of `forceRefresh` as the bypass flag would let followers use the cache. Tiny optimization, not urgent.

gabriel-adopt:
Thanks — all three addressed in `e46c016`:

1. **Header limit authoritative** — added a post-merge `MAX_HEADER_COUNT` check in `executeFetch` (throws 400) so merging captured headers can't silently exceed the limit; unit test added.
2. **Silent no-op without `credential_types.headers`** — the empty-capture path now logs a louder warning that names the missing field (and distinguishes "not declared" vs "bundle has none yet"), and the Swagger description now states the prerequisite.
3. **Non-leader cache bypass** — `getCredentialsForSession` now bypasses the cache only for the coalesce leader; followers reuse the cache instead of a redundant MinIO round-trip.

Build + lint green, 586 API tests pass (+1 new).

gabriel-adopt:
Second round addressed in `9b414c1`:

1. **refresh doesn't wait** — `executeFetch` now passes `waitSeconds=5` when `refresh_credentials` is set, so it blocks (best-effort) for `extractDone` before reading, instead of returning the cached bundle. Matches the DTO contract.

2. **Cookie filter** — `mergeCapturedHeaders` now actually skips `cookie`/`cookie2` (and defends the CSRF header name), so a misconfigured capture can't make the in-page fetch throw. Added a unit test.

3. **Non-leader cache vs. fresh** — this one was in tension with an earlier maintainer note (followers should reuse cache to avoid a redundant MinIO round-trip). Reconciled by keying off `waitSeconds`:
   - `waitSeconds > 0` (the refresh path, now always the case from #1): **every** caller — leader or concurrent follower — waits for `extractDone` and bypasses the cache, so nobody gets a stale bundle (your point).
   - `waitSeconds == 0`: fire-and-forget — only the leader bypasses; followers reuse cache (the earlier optimization).

   So a real refresh is fresh for all callers, while fire-and-forget stays cheap.

588 API tests pass (build + lint green).

joaomoraski-adoptai (approved):

PR #147, opened 2026-07-30
fix: improve local dev experience — bootstrap, egress proxy, Swagger docs, Kind MTU

navjot-adopt:
## Summary

Addresses 5 issues from the [local dev setup retro](https://claude.ai/code/artifact/5651af00-adcb-42a3-8c5b-c047cd2845f9) that every new developer hits when setting up Tabby + NoUI in a Kind cluster.

### Issue #3 — Bootstrap skipped by phantom tenant
- The recording pool system tenant (`__system_recording_pool__`, seeded by migration 033) was counted by `bootstrap.service.ts`, preventing auto-provisioning of the admin user on fresh installs
- **Fix:** Exclude `RECORDING_POOL.SYSTEM_TENANT_ID` from the tenant count so bootstrap runs when only the system tenant exists

### Issue #5 — App creation has undocumented required fields
- Swagger `@ApiProperty` examples showed incomplete nested payloads for `login_config`, `keepalive_config`, `export_policy`, and `notification_config`, forcing trial-and-error 400 responses
- **Fix:** Updated all examples to show minimal valid payloads and added descriptions listing every required sub-field and its constraints

### Issue #8 — NetworkPolicy blocks all worker egress in local dev
- Controller creates a `np-{sessionId}` NetworkPolicy restricting worker pod egress. Must be manually deleted for local dev
- **Fix:** Added `DISABLE_NETWORK_POLICY` config flag (default `"false"`, set to `"true"` in `values-local.yaml`). When enabled, skips K8s NP creation and forces `allow_all=true` on the egress proxy. No effect on staging/prod (flag not in deploy pipeline)

### Issue #9 — Egress proxy silent socket close on CONNECT error
- The upstream socket error handler silently destroyed the client socket with no HTTP response, producing confusing "connection ended before receiving CONNECT response" errors in Playwright
- **Fix:** Now writes `HTTP/1.1 502 Bad Gateway` with `X-Proxy-Error` header before destroying the socket

### Kind MTU fix
- kindnet CNI defaults to MTU 65535, silently breaking TLS handshakes to external sites (GitHub, LinkedIn CDN) whose certificates produce large packets
- **Fix:** Added `kind-fix-mtu` Makefile target + `infra/kind/cluster-config.yaml`. `make kind-create` now auto-patches MTU to 1500

### Kind CoreDNS fix (added in d735bbf)
- Docker Desktop's embedded DNS (which Kind's CoreDNS forwards to via `/etc/resolv.conf`) returns SERVFAIL on AAAA lookups for CNAME→CloudFront domains. `getaddrinfo` (Chromium + the egress proxy's `net.connect`) does a dual A+AAAA lookup and fails the whole resolution with EAI_AGAIN → `ERR_TUNNEL_CONNECTION_FAILED` in worker sessions
- **Fix:** Added `kind-fix-dns` Makefile target that repoints CoreDNS `forward` at `8.8.8.8 1.1.1.1`, wired into `make kind-create`. Both `kind-fix-*` targets pin `--context kind-<cluster>` + a guard so they can never mutate a remote cluster, and verify-then-report (fail loud on no-op). Local-dev only — not in `values-*`, TFY, or CI

### CloakBrowser upgrade (0.3.12 → 0.5.2)
- v0.3.12 had no macOS/ARM64 binary; worker fell back to stock Playwright which triggers bot detection on LinkedIn
- **Fix:** Bumped to v0.5.2 which supports all platforms

## Staging/prod safety
- `DISABLE_NETWORK_POLICY` defaults to `"false"` in `values.yaml` and is **not** referenced in `infra/tfy/deploy.yaml` or any GitHub Actions workflow — staging/prod always get the default
- Bootstrap change only affects fresh installs where only the system tenant exists
- Swagger example changes are documentation-only
- Egress proxy 502 response is strictly more informative than the previous silent close

## Files changed
| File | What |
|------|------|
| `apps/api/src/modules/auth/bootstrap.service.ts` | Exclude system tenant from bootstrap count |
| `apps/api/src/modules/apps/apps.dto.ts` | Improved Swagger examples with minimal valid payloads |
| `apps/controller/src/reconcile.service.ts` | `DISABLE_NETWORK_POLICY` flag support |
| `charts/browser-hitl/files/egress-proxy/server.js` | 502 response on CONNECT upstream error |
| `charts/browser-hitl/values.yaml` | New `disableNetworkPolicy` config |
| `charts/browser-hitl/values-local.yaml` | Enable `disableNetworkPolicy` for local dev |
| `charts/browser-hitl/templates/configmap.yaml` | Render new config into ConfigMap |
| `Makefile` | `kind-fix-mtu` + `kind-fix-dns` targets (context-pinned + verify-then-report), updated `kind-create` |
| `infra/kind/cluster-config.yaml` | Minimal Kind config (MTU patched post-create by `kind-fix-mtu`) |
| `apps/worker/package.json` + `pnpm-lock.yaml` | CloakBrowser 0.3.12 → 0.5.2 |
| `CLAUDE.md` | Document `kind-fix-mtu`/`kind-fix-dns` + MTU/CoreDNS gotchas |

## Test plan
- [x] `pnpm run build` — all 7 projects pass
- [x] `pnpm run test` — all 601 tests pass
- [x] `helm lint charts/browser-hitl/` — passes
- [x] Verify bootstrap provisions admin user on fresh Kind install (delete PVC + pods, redeploy)
- [x] Verify `DISABLE_NETWORK_POLICY=true` skips NP creation and pushes `allow_all` (check controller logs)
- [x] Verify egress proxy returns 502 on upstream connect failure (curl CONNECT to unreachable host)
- [x] Verify Swagger UI shows improved examples at `/api/docs`
- [x] **CloakBrowser 0.5.2 proof** — `make docker-build-worker` succeeds (exit 0; Dockerfile runs `ensureBinary()`); `docker run … worker:dev` confirms `ensureBinary: function | launch: function` and `/opt/cloakbrowser` holds the chromium binaries; a live worker session logged `Browser launched via CloakBrowser (stealth mode)`
- [x] `kind-fix-dns`/`kind-fix-mtu` verified: run against `kind-tabby-dev`, guard refuses a non-existent context, verify-then-report passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,8 +240,21 @@ tilt: ## Start Tilt for live rebuild and deploy (replaces kind-reload-all)
 
 .PHONY: kind-create
 kind-create: ## Create a Kind cluster for local development
-	kind create cluster --name $(KIND_CLUSTER)
+	kind create cluster --name $(KIND_CLUSTER) --config infra/kind/cluster-config.yaml
+	$(MAKE) kind-fix-mtu
 	@echo "Kind cluster '$(KIND_CLUSTER)' created. Context: kind-$(KIND_CLUSTER)"
+
+.PHONY: kind-fix-mtu
+kind-fix-mtu: ## Fix Kind pod MTU from 65535 to 1500 (prevents TLS failures to external sites)
+	@echo "Patching kindnet MTU to 1500..."
+	@docker exec $(KIND_CLUSTER)-control-plane sh -c '\
+		CNI=/etc/cni/net.d/10-kindnet.conflist; \
+		if [ -s "$$CNI" ]; then \
+			sed -i "s/\"mtu\": *[0-9]*/\"mtu\": 1500/g" "$$CNI"; \
+		fi'
+	kubectl rollout restart daemonset/kindnet -n kube-system
+	kubectl rollout status daemonset/kindnet -n kube-system --timeout=30s
+	@echo "MTU fixed to 1500. Restart pods to pick up the new MTU."
 
 .PHONY: kind-load-images
 kind-load-images: ## Load all Docker images into the Kind cluster
--- a/Makefile
+++ b/Makefile
@@ -245,33 +245,54 @@ kind-create: ## Create a Kind cluster for local development
 	$(MAKE) kind-fix-dns
 	@echo "Kind cluster '$(KIND_CLUSTER)' created. Context: kind-$(KIND_CLUSTER)"
 
+# The Kind cluster's kubectl context (Kind names it kind-<cluster>). All mutating
+# targets below pin --context to this so they can NEVER touch a remote cluster,
+# even if the current kube-context is pointed at staging/prod.
+KIND_CONTEXT := kind-$(KIND_CLUSTER)
+
+# Guard: refuse to run if the Kind cluster's context doesn't exist (i.e. the
+# cluster isn't up). Prevents these local-only targets from silently no-oping
+# against whatever context happens to be current.
+.PHONY: kind-guard
+kind-guard:
+	@kubectl config get-contexts -o name 2>/dev/null | grep -qx '$(KIND_CONTEXT)' || \
+		{ echo "refusing: kube-context '$(KIND_CONTEXT)' not found — is the Kind cluster up? (make kind-create)"; exit 1; }
+
 .PHONY: kind-fix-mtu
-kind-fix-mtu: ## Fix Kind pod MTU from 65535 to 1500 (prevents TLS failures to external sites)
-	@echo "Patching kindnet MTU to 1500..."
-	@docker exec $(KIND_CLUSTER)-control-plane sh -c '\
-		CNI=/etc/cni/net.d/10-kindnet.conflist; \
-		if [ -s "$$CNI" ]; then \
-			sed -i "s/\"mtu\": *[0-9]*/\"mtu\": 1500/g" "$$CNI"; \
-		fi'
-	kubectl rollout restart daemonset/kindnet -n kube-system
-	kubectl rollout status daemonset/kindnet -n kube-system --timeout=30s
-	@echo "MTU fixed to 1500. Restart pods to pick up the new MTU."
+kind-fix-mtu: kind-guard ## Fix Kind pod MTU from 65535 to 1500 (prevents TLS failures to external sites)
+	@echo "Patching kindnet MTU to 1500 on all nodes of '$(KIND_CLUSTER)'..."
+	@# kindnet rewrites its conflist on startup, so restart the daemonset FIRST,
+	@# then patch every node's conflist, so the patch is what survives.
+	kubectl --context $(KIND_CONTEXT) rollout restart daemonset/kindnet -n kube-system
+	kubectl --context $(KIND_CONTEXT) rollout status daemonset/kindnet -n kube-system --timeout=30s
+	@for node in $$(kind get nodes --name $(KIND_CLUSTER)); do \
+		docker exec $$node sh -c '\
+			CNI=/etc/cni/net.d/10-kindnet.conflist; \
+			[ -s "$$CNI" ] && sed -i "s/\"mtu\": *[0-9]*/\"mtu\": 1500/g" "$$CNI"; \
+			grep -q "\"mtu\": 1500" "$$CNI"' \
+			|| { echo "MTU patch did not land on $$node (no \"mtu\" key? kindnet variant changed)"; exit 1; }; \
+	done
+	@echo "MTU set to 1500 on all nodes (verified). Restart pods to pick it up."
 
 .PHONY: kind-fix-dns
-kind-fix-dns: ## Point CoreDNS at public resolvers (fixes AAAA SERVFAIL from Docker Desktop embedded DNS)
-	@echo "Pointing CoreDNS forward at 8.8.8.8 1.1.1.1 (local-dev only)..."
+kind-fix-dns: kind-guard ## Point CoreDNS at public resolvers (fixes AAAA SERVFAIL from Docker Desktop embedded DNS)
+	@echo "Pointing CoreDNS forward at 8.8.8.8 1.1.1.1 on '$(KIND_CONTEXT)' (local-dev only)..."
 	@# Docker Desktop's embedded DNS (127.0.0.11, reached via /etc/resolv.conf) returns
 	@# SERVFAIL on AAAA lookups for CNAME->CloudFront domains. getaddrinfo (Chromium + the
 	@# egress proxy's net.connect) does a dual A+AAAA lookup and fails the whole resolution
 	@# with EAI_AGAIN, surfacing as ERR_TUNNEL_CONNECTION_FAILED in worker sessions.
 	@# Forwarding to a public resolver that answers AAAA cleanly avoids this. Cloud clusters
 	@# (staging/prod) use a well-behaved VPC resolver and never hit this, so it stays local.
-	@kubectl get configmap coredns -n kube-system -o yaml | \
+	@kubectl --context $(KIND_CONTEXT) get configmap coredns -n kube-system -o yaml | \
 		sed 's#forward . /etc/resolv.conf#forward . 8.8.8.8 1.1.1.1#' | \
-		kubectl apply -f -
-	kubectl rollout restart deployment/coredns -n kube-system
-	kubectl rollout status deployment/coredns -n kube-system --timeout=60s
-	@echo "CoreDNS now forwards to public resolvers."
+		kubectl --context $(KIND_CONTEXT) apply -f -
+	@# Verify the substitution actually landed (CoreDNS Corefile variants differ; a
+	@# no-op sed would otherwise re-apply an identical ConfigMap and still report success).
+	@kubectl --context $(KIND_CONTEXT) get configmap coredns -n kube-system -o yaml | grep -q '8.8.8.8' \
+		|| { echo "CoreDNS Corefile did not match 'forward . /etc/resolv.conf' — nothing changed. Inspect the Corefile manually."; exit 1; }
+	kubectl --context $(KIND_CONTEXT) rollout restart deployment/coredns -n kube-system
+	kubectl --context $(KIND_CONTEXT) rollout status deployment/coredns -n kube-system --timeout=60s
+	@echo "CoreDNS now forwards to public resolvers (verified 8.8.8.8 in Corefile)."
 
 .PHONY: kind-load-images
 kind-load-images: ## Load all Docker images into the Kind cluster
--- a/Makefile
+++ b/Makefile
@@ -242,6 +242,7 @@ tilt: ## Start Tilt for live rebuild and deploy (replaces kind-reload-all)
 kind-create: ## Create a Kind cluster for local development
 	kind create cluster --name $(KIND_CLUSTER) --config infra/kind/cluster-config.yaml
 	$(MAKE) kind-fix-mtu
+	$(MAKE) kind-fix-dns
 	@echo "Kind cluster '$(KIND_CLUSTER)' created. Context: kind-$(KIND_CLUSTER)"
 
 .PHONY: kind-fix-mtu
@@ -255,6 +256,22 @@ kind-fix-mtu: ## Fix Kind pod MTU from 65535 to 1500 (prevents TLS failures to e
 	kubectl rollout restart daemonset/kindnet -n kube-system
 	kubectl rollout status daemonset/kindnet -n kube-system --timeout=30s
 	@echo "MTU fixed to 1500. Restart pods to pick up the new MTU."
+
+.PHONY: kind-fix-dns
+kind-fix-dns: ## Point CoreDNS at public resolvers (fixes AAAA SERVFAIL from Docker Desktop embedded DNS)
+	@echo "Pointing CoreDNS forward at 8.8.8.8 1.1.1.1 (local-dev only)..."
+	@# Docker Desktop's embedded DNS (127.0.0.11, reached via /etc/resolv.conf) returns
+	@# SERVFAIL on AAAA lookups for CNAME->CloudFront domains. getaddrinfo (Chromium + the
+	@# egress proxy's net.connect) does a dual A+AAAA lookup and fails the whole resolution
+	@# with EAI_AGAIN, surfacing as ERR_TUNNEL_CONNECTION_FAILED in worker sessions.
+	@# Forwarding to a public resolver that answers AAAA cleanly avoids this. Cloud clusters
+	@# (staging/prod) use a well-behaved VPC resolver and never hit this, so it stays local.
+	@kubectl get configmap coredns -n kube-system -o yaml | \
+		sed 's#forward . /etc/resolv.conf#forward . 8.8.8.8 1.1.1.1#' | \
+		kubectl apply -f -
+	kubectl rollout restart deployment/coredns -n kube-system
+	kubectl rollout status deployment/coredns -n kube-system --timeout=60s
+	@echo "CoreDNS now forwards to public resolvers."
 
 .PHONY: kind-load-images
 kind-load-images: ## Load all Docker images into the Kind cluster
--- a/apps/api/src/modules/apps/apps.dto.ts
+++ b/apps/api/src/modules/apps/apps.dto.ts
@@ -56,19 +56,46 @@ export class CreateAppDto {
   @IsString({ each: true })
   extra_egress_allowlist?: string[];
 
-  @ApiProperty({ description: 'Login DSL configuration with URL, credential ref, and steps', example: { login_url: 'https://app.hubspot.com/login', credential_ref: 'k8s:secret/hubspot-creds', steps: [{ action: 'goto', url: 'https://app.hubspot.com/login' }, { action: 'fill', selector: '#username', value: '${USERNAME}' }, { action: 'click', selector: '#loginBtn' }] } })
+  @ApiProperty({
+    description: 'Login DSL configuration. Required fields: login_url, credential_ref (k8s:secret/{name} or manual:), steps (array with at least one goto action). Step actions: goto, fill, type, click, select, wait_for, wait_for_url, frame, main_frame, popup, keyboard, evaluate, sleep, screenshot, reload, request_human_input.',
+    example: {
+      login_url: 'https://app.hubspot.com/login',
+      credential_ref: 'k8s:secret/hubspot-creds',
+      steps: [
+        { action: 'goto', url: 'https://app.hubspot.com/login' },
+        { action: 'fill', selector: '#username', value: '${USERNAME}' },
+        { action: 'fill', selector: '#password', value: '${PASSWORD}' },
+        { action: 'click', selector: '#loginBtn' },
+      ],
+    },
+  })
   @IsObject()
   login_config: Record<string, unknown>;
 
-  @ApiProperty({ description: 'Health check configuration', example: { interval_seconds: 300, actions: [{ action: 'goto', url: 'https://app.hubspot.com/home' }], health_checks: [{ type: 'url_check', url: 'https://app.hubspot.com/home', expect_status: 200 }], policy: 'all' } })
+  @ApiProperty({
+    description: 'Health check configuration. Required fields: interval_seconds (>= 60), actions (array, can be empty), health_checks (non-empty array). Health check types: url_check (needs url + expect_status), dom_check (needs selector), network_check (needs url + expect_status). policy: all | any | quorum.',
+    example: {
+      interval_seconds: 300,
+      actions: [],
+      health_checks: [{ type: 'url_check', url: 'https://app.hubspot.com/home', expect_status: 200 }],
+      policy: 'all',
+    },
+  })
   @IsObject()
   keepalive_config: Record<string, unknown>;
 
-  @ApiProperty({ description: 'Credential export configuration', example: { artifact_types: ['cookies', 'headers', 'csrf_token'], encryption: { algo: 'AES-256-GCM', key_ref: 'k8s:secret/tenant-key' }, ttl_seconds: 3600 } })
+  @ApiProperty({
+    description: 'Credential export configuration. Required fields: artifact_types (non-empty array from: cookies, headers, csrf_token, local_storage, session_storage), encryption (algo must be AES-256-GCM), ttl_seconds (>= 300).',
+    example: {
+      artifact_types: ['cookies'],
+      encryption: { algo: 'AES-256-GCM' },
+      ttl_seconds: 3600,
+    },
+  })
   @IsObject()
   export_policy: Record<string, unknown>;
 
-  @ApiProperty({ description: 'Notification channels for HITL events. Omit for silent/agent-poll mode.', example: { channels: ['slack:#tabby-experiments'] }, required: false })
+  @ApiProperty({ description: 'Notification channels for HITL events. Omit entirely for silent/agent-poll mode. Channels format: {provider}:{reference} where provider is slack, teams, or agent.', example: { channels: [] }, required: false })
   @IsOptional()
   @IsObject()
   notification_config?: Record<string, unknown>;
--- a/apps/api/src/modules/apps/apps.dto.ts
+++ b/apps/api/src/modules/apps/apps.dto.ts
@@ -73,7 +73,7 @@ export class CreateAppDto {
   login_config: Record<string, unknown>;
 
   @ApiProperty({
-    description: 'Health check configuration. Required fields: interval_seconds (>= 60), actions (array, can be empty), health_checks (non-empty array). Health check types: url_check (needs url + expect_status), dom_check (needs selector), network_check (needs url + expect_status). policy: all | any | quorum.',
+    description: 'Health check configuration. Required fields: interval_seconds (>= 60), actions (array, can be empty), health_checks (non-empty array). Health check types: url_check (needs url + expect_status), dom_check (needs selector), network_check (needs url + expect_status). policy: all | any | quorum (quorum requires quorum_n >= 1).',
     example: {
       interval_seconds: 300,
       actions: [],
@@ -85,10 +85,10 @@ export class CreateAppDto {
   keepalive_config: Record<string, unknown>;
 
   @ApiProperty({
-    description: 'Credential export configuration. Required fields: artifact_types (non-empty array from: cookies, headers, csrf_token, local_storage, session_storage), encryption (algo must be AES-256-GCM), ttl_seconds (>= 300).',
+    description: 'Credential export configuration. Required fields: artifact_types (non-empty array from: cookies, headers, csrf_token, local_storage, session_storage), encryption (algo must be AES-256-GCM, key_ref format: k8s:secret/{name}), ttl_seconds (>= 300).',
     example: {
       artifact_types: ['cookies'],
-      encryption: { algo: 'AES-256-GCM' },
+      encryption: { algo: 'AES-256-GCM', key_ref: 'k8s:secret/tenant-key' },
       ttl_seconds: 3600,
     },
   })
--- a/apps/api/src/modules/auth/bootstrap.service.ts
+++ b/apps/api/src/modules/auth/bootstrap.service.ts
@@ -1,7 +1,8 @@
 import { Injectable, OnModuleInit, Logger } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { randomUUID } from 'crypto';
-import { Repository } from 'typeorm';
+import { Not, Repository } from 'typeorm';
+import { RECORDING_POOL } from '@browser-hitl/shared';
 import { TenantEntity, UserEntity } from '../../entities';
 import { AuthService } from './auth.service';
 
@@ -26,7 +27,9 @@ export class BootstrapService implements OnModuleInit {
   ) {}
 
   async onModuleInit() {
-    const tenantCount = await this.tenantRepo.count();
+    const tenantCount = await this.tenantRepo.count({
+      where: { id: Not(RECORDING_POOL.SYSTEM_TENANT_ID) },
+    });
     if (tenantCount > 0) {
       this.logger.log('Tenants exist, skipping bootstrap');
       return;
--- a/apps/api/src/modules/auth/bootstrap.service.ts
+++ b/apps/api/src/modules/auth/bootstrap.service.ts
@@ -12,7 +12,9 @@ import { AuthService } from './auth.service';
  * 1. A tenant with name BOOTSTRAP_TENANT_NAME
  * 2. An admin user with ADMIN_BOOTSTRAP_EMAIL/PASSWORD
  * 3. (MinIO bucket + encryption key provisioned separately)
- * Idempotent: skips if any tenant exists.
+ * Idempotent: skips if any *non-system* tenant exists. The system recording-pool
+ * tenant (RECORDING_POOL.SYSTEM_TENANT_ID, seeded by migration 033) is excluded
+ * from the count, so a fresh install with only the system tenant still bootstraps.
  */
 @Injectable()
 export class BootstrapService implements OnModuleInit {
--- a/apps/controller/src/reconcile.service.spec.ts
+++ b/apps/controller/src/reconcile.service.spec.ts
@@ -285,7 +285,9 @@ describe('ReconcileService DISABLE_NETWORK_POLICY', () => {
       state_version: 1,
       retry_count: 0,
       owner_user_id: null,
-      residential_proxy_override: null,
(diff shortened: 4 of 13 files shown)


navjot-adopt:
## Test Plan Results ✓

All 4 pending test plan items verified on a local Kind cluster:

### 1. Bootstrap provisions admin on fresh install ✓
- Deleted postgres PVC + pod to reset DB
- After redeploy, API logs show:
  ```
  [BootstrapService] Bootstrapping tenant: default, admin: admin@browser-hitl.local
  [BootstrapService] Bootstrap complete: tenant=cf94c0a7-..., admin=15d1f361-...
  ```
- Migration 033 seeds the `__system_recording_pool__` tenant **before** bootstrap runs — previously this blocked bootstrap. Now correctly excluded.
- Login with bootstrap password (`LocalDev123!@#`) succeeds immediately — no manual bcrypt hash generation needed.

### 2. DISABLE_NETWORK_POLICY=true skips NP creation ✓
- ConfigMap confirmed: `DISABLE_NETWORK_POLICY: "true"`
- Controller logs for new session: `Skipped NetworkPolicy (DISABLE_NETWORK_POLICY=true), pushed allow_all for session 5fe03067-...`
- `kubectl get networkpolicies -n browser-hitl` → `No resources found`
- Egress proxy allowlist for the session has `["*", "www.linkedin.com"]` (allow_all applied)

### 3. Egress proxy returns 502 on CONNECT error ✓
- `curl --proxy ... https://127.0.0.1:1/` → refused connection immediately returns:
  ```
  HTTP/1.1 502 Bad Gateway
  X-Proxy-Error: connect ECONNREFUSED 127.0.0.1:1
  ```
- Previously: silent socket close with no HTTP response → confusing "connection ended before receiving CONNECT response" in Playwright

### 4. Swagger UI shows improved examples ✓
- `/api/docs-json` confirms updated `@ApiProperty` descriptions and examples:
  - `keepalive_config`: shows `actions: []` (minimal valid)
  - `export_policy`: shows `artifact_types: ["cookies"], encryption: {algo: "AES-256-GCM"}, ttl_seconds: 3600`
  - `notification_config`: shows `channels: []` with description noting "Omit entirely for silent mode"
  - Descriptions now list all required sub-fields and constraints

charts/browser-hitl/files/egress-proxy/server.js
```diff
@@ -703,7 +703,12 @@ function proxyConnect(req, clientSocket, head) {
     clientSocket.pipe(upstreamSocket);
   });
 
-  upstreamSocket.on('error', () => {
+  upstreamSocket.on('error', (err) => {
+    if (!clientSocket.destroyed) {
+      try {
+        clientSocket.write(`HTTP/1.1 502 Bad Gateway\r\nX-Proxy-Error: ${String(err?.message || 'upstream connect failed').replace(/[\r\n]/g, ' ')}\r\nConnection: close\r\n\r\n`);
+      } catch {}
```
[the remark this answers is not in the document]
navjot-adopt: Fixed in b8e2cc4 — added `tunnelEstablished` flag. The 502 response is now only written when `!tunnelEstablished && !clientSocket.destroyed`, preventing bytes from being injected into an established TLS stream.

apps/controller/src/reconcile.service.ts
```diff
@@ -388,9 +391,13 @@ export class ReconcileService implements OnModuleInit, OnModuleDestroy {
...
       }
 
-      // Generate NetworkPolicy — open the 8091 ingress when the worker health
-      // Service exists (execute or recording drain).
-      await this.podManager.createNetworkPolicy(savedSession.id, podName, app.target_urls, streamingMode, needsWorkerHealth, extraAllowlist, allowAll, this.resolveResidential(savedSession, app));
+      if (this.disableNetworkPolicy) {
+        // Local dev: skip K8s NetworkPolicy, push allow_all to the egress proxy
+        await this.podManager.syncEgressAllowlist(savedSession.id, app.target_urls, extraAllowlist, true, this.resolveResidential(savedSession, app));
+        this.logger.log(`Skipped NetworkPolicy (DISABLE_NETWORK_POLICY=true), pushed allow_all for session ${savedSession.id}`);
+      } else {
```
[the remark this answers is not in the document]
navjot-adopt: Fixed in b8e2cc4 — added two unit tests in `reconcile.service.spec.ts`: one verifying `createNetworkPolicy` is skipped and `syncEgressAllowlist` is called with `allowAll=true` when the flag is set, and another confirming normal `createNetworkPolicy` path when the flag is unset.

apps/api/src/modules/apps/apps.dto.ts
```diff
@@ -56,19 +56,46 @@ export class CreateAppDto {
...
   @IsObject()
   keepalive_config: Record<string, unknown>;
 
-  @ApiProperty({ description: 'Credential export configuration', example: { artifact_types: ['cookies', 'headers', 'csrf_token'], encryption: { algo: 'AES-256-GCM', key_ref: 'k8s:secret/tenant-key' }, ttl_seconds: 3600 } })
+  @ApiProperty({
+    description: 'Credential export configuration. Required fields: artifact_types (non-empty array from: cookies, headers, csrf_token, local_storage, session_storage), encryption (algo must be AES-256-GCM), ttl_seconds (>= 300).',
+    example: {
+      artifact_types: ['cookies'],
+      encryption: { algo: 'AES-256-GCM' },
+      ttl_seconds: 3600,
```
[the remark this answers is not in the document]
navjot-adopt: Fixed in b8e2cc4 — added `key_ref: 'k8s:secret/tenant-key'` to the encryption example and updated the description to mention `key_ref format: k8s:secret/{name}`.

apps/api/src/modules/apps/apps.dto.ts
```diff
@@ -56,19 +56,46 @@ export class CreateAppDto {
...
+      ],
+    },
+  })
   @IsObject()
   login_config: Record<string, unknown>;
 
-  @ApiProperty({ description: 'Health check configuration', example: { interval_seconds: 300, actions: [{ action: 'goto', url: 'https://app.hubspot.com/home' }], health_checks: [{ type: 'url_check', url: 'https://app.hubspot.com/home', expect_status: 200 }], policy: 'all' } })
+  @ApiProperty({
+    description: 'Health check configuration. Required fields: interval_seconds (>= 60), actions (array, can be empty), health_checks (non-empty array). Health check types: url_check (needs url + expect_status), dom_check (needs selector), network_check (needs url + expect_status). policy: all | any | quorum.',
+    example: {
```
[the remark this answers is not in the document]
navjot-adopt: Fixed in b8e2cc4 — added "quorum requires quorum_n >= 1" to the keepalive_config description. Regarding dom_check `exists` boolean: the validator doesn't enforce it and the worker defaults to `isVisible()` when omitted, so it's not a required field in the Swagger example.

gabriel-adopt:
Reviewing it @navjot-adopt

apps/worker/package.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": {
     "@browser-hitl/shared": "workspace:*",
     "@sentry/node": "^10.51.0",
-    "cloakbrowser": "^0.3.12",
+    "cloakbrowser": "^0.5.2",
```
gabriel-adopt, line 14: **This is the gating item — the only prod-affecting change in the PR, and the one with no proof of change.**

`0.3.12 → 0.5.2` on a `0.x` package crosses two breaking-change vehicles, and three contracts here still depend on the old API surface:

- `infra/docker/Dockerfile.worker:40` imports the named export `ensureBinary()`
- `infra/docker/Dockerfile.worker:39` relies on `CLOAKBROWSER_CACHE_DIR` being honored
- `apps/worker/src/main.ts:167` imports the named export `launch()`

The lockfile shows 0.5.2 gained `hasBin: true` — a CLI is exactly where binary-provisioning APIs tend to migrate. If `ensureBinary` moved or the cache layout changed, the Docker build either fails (loud, fine) or the binary lands somewhere the runtime doesn't look — and `main.ts:170` catches that into a bare `console.warn('CloakBrowser unavailable, falling back to stock Playwright')`, silently reinstating the LinkedIn bot detection this bump exists to fix.

**CI cannot catch this**: `ci.yaml` has lint / test / build / audit / helm-lint jobs and no Docker image build, so the green checkmarks say nothing about `ensureBinary`. Could you add to the test plan: `make docker-build-worker` succeeds, plus one real session whose log reads `Browser launched via CloakBrowser (stealth mode)`?

Separate follow-up worth filing: that fallback should be a Sentry capture, not a `console.warn` — the worker already depends on `@sentry/node`, and a silent stealth downgrade in prod is precisely the class of bug that cost this repo days on the residential path.

Two smaller notes on the same bump, neither blocking: the new optional peer `socks-proxy-agent >=10.0.0` is unmet (fine today — residential egress is HTTP CONNECT through the egress proxy, and nothing in `apps/worker/src` or `server.js` references SOCKS); and the lockfile picked up `libc:` fields across sharp/next/nx, meaning it was regenerated by a pnpm newer than the `packageManager: pnpm@10.12.1` pin. Harmless (unknown keys are tolerated under `--frozen-lockfile`) but it's unrelated churn in the diff — either bump the pin or regenerate with 10.12.1.

charts/browser-hitl/files/egress-proxy/server.js
```diff
@@ -694,16 +694,24 @@ function proxyConnect(req, clientSocket, head) {
...
     if (head && head.length > 0) {
       upstreamSocket.write(head);
     }
     upstreamSocket.pipe(clientSocket);
     clientSocket.pipe(upstreamSocket);
   });
 
-  upstreamSocket.on('error', () => {
+  upstreamSocket.on('error', (err) => {
+    if (!tunnelEstablished && !clientSocket.destroyed) {
```
gabriel-adopt, line 712: The `tunnelEstablished` guard is right, and surfacing the reason to the client is a real improvement. But this now tells the *client* and still tells the *logs* nothing.

Compare the residential sibling `fail()` at `server.js:519-527`: it logs `residential CONNECT failed` with `session_id`, target, and reason **before** writing its 502. Given this repo just spent three stacked silent bugs on the residential path — one of which was precisely "no log on a declined decision" — the log line is the half that saves the next debugging session. `sessionId` is already in scope here:

```js
upstreamSocket.on('error', (err) => {
  const reason = String(err?.message || 'upstream connect failed');
  if (!tunnelEstablished && !clientSocket.destroyed) {
    log('direct CONNECT failed', { session_id: sessionId, target: `${hostname}:${port}`, reason });
    // ...write 502
  }
  clientSocket.destroy();
});
```

Second, smaller point: `denyConnect` (:414), `requireProxyAuth` (:481), and `fail()` (:519) all send a JSON body with `Content-Type` and `Content-Length`; this response is headers-only, with the detail in a custom `X-Proxy-Error` header instead. The `[\r\n]` scrub is the right instinct, but matching the established body shape costs two lines and keeps the proxy's error surface uniform for whoever writes the next client against it.

Makefile
```diff
@@ -240,9 +240,39 @@ tilt: ## Start Tilt for live rebuild and deploy (replaces kind-reload-all)
...
+	@echo "Pointing CoreDNS forward at 8.8.8.8 1.1.1.1 (local-dev only)..."
+	@# Docker Desktop's embedded DNS (127.0.0.11, reached via /etc/resolv.conf) returns
+	@# SERVFAIL on AAAA lookups for CNAME->CloudFront domains. getaddrinfo (Chromium + the
+	@# egress proxy's net.connect) does a dual A+AAAA lookup and fails the whole resolution
+	@# with EAI_AGAIN, surfacing as ERR_TUNNEL_CONNECTION_FAILED in worker sessions.
+	@# Forwarding to a public resolver that answers AAAA cleanly avoids this. Cloud clusters
+	@# (staging/prod) use a well-behaved VPC resolver and never hit this, so it stays local.
+	@kubectl get configmap coredns -n kube-system -o yaml | \
+		sed 's#forward . /etc/resolv.conf#forward . 8.8.8.8 1.1.1.1#' | \
+		kubectl apply -f -
```
gabriel-adopt: **This will happily rewrite a remote cluster's CoreDNS.** Every command in this target runs against whatever kube-context happens to be current — there's no Kind scoping at all. Run `make kind-fix-dns` with staging selected and you repoint *that* cluster's CoreDNS at `8.8.8.8`/`1.1.1.1` and restart its deployment.

`kind-fix-mtu` at least opens with `docker exec $(KIND_CLUSTER)-control-plane`, which fails fast off-Kind — though its two `kubectl rollout` calls have the same exposure once execution gets that far.

Suggest `--context kind-$(KIND_CLUSTER)` on all four kubectl calls across both targets (Kind names the context `kind-<cluster>`), or an early guard:

```make
@kubectl config current-context | grep -q '^kind-' || { echo "refusing: current context is not a kind cluster"; exit 1; }
```

I know the Makefile has no `--context` precedent — `K8S_CONTEXT` at :34 just reads the current one — but no existing target is capable of mutating a remote cluster's DNS either, so the precedent doesn't really cover this case.

Makefile
```diff
@@ -240,9 +240,39 @@ tilt: ## Start Tilt for live rebuild and deploy (replaces kind-reload-all)
...
+	kubectl rollout restart deployment/coredns -n kube-system
+	kubectl rollout status deployment/coredns -n kube-system --timeout=60s
+	@echo "CoreDNS now forwards to public resolvers."
```
gabriel-adopt: This prints unconditionally, including when nothing changed. If the Corefile doesn't match `forward . /etc/resolv.conf` (CoreDNS variants differ), the `sed` is a no-op, `kubectl apply` re-applies an identical ConfigMap, and the target still announces "CoreDNS now forwards to public resolvers."

A silent no-op that reports success is the exact failure mode the retro was written about. Verify before claiming — e.g. capture the rendered YAML, `grep -q '8.8.8.8'`, and fail loudly otherwise.

Makefile
```diff
@@ -240,9 +240,39 @@ tilt: ## Start Tilt for live rebuild and deploy (replaces kind-reload-all)
...
+kind-fix-mtu: ## Fix Kind pod MTU from 65535 to 1500 (prevents TLS failures to external sites)
+	@echo "Patching kindnet MTU to 1500..."
+	@docker exec $(KIND_CLUSTER)-control-plane sh -c '\
+		CNI=/etc/cni/net.d/10-kindnet.conflist; \
+		if [ -s "$$CNI" ]; then \
+			sed -i "s/\"mtu\": *[0-9]*/\"mtu\": 1500/g" "$$CNI"; \
+		fi'
+	kubectl rollout restart daemonset/kindnet -n kube-system
+	kubectl rollout status daemonset/kindnet -n kube-system --timeout=30s
+	@echo "MTU fixed to 1500. Restart pods to pick up the new MTU."
```
gabriel-adopt: Same verify-then-report issue as `kind-fix-dns`, plus two mechanics worth a second look.

1. **Silent no-op.** If the conflist has no `"mtu"` key (this varies by kindnet version) or the `[ -s "$$CNI" ]` test fails, the `sed` changes nothing and this line still says "MTU fixed to 1500." Grep the patched file and fail loudly if the substitution didn't land.
2. **Ordering.** :254 patches `10-kindnet.conflist`, then :256 does `rollout restart daemonset/kindnet`. Kindnet rewrites that conflist at startup in several versions — which would mean the restart undoes the patch you just applied and MTU stays 65535. If you verified this against a specific kindnet version, please pin the version in a comment; otherwise the restart needs to come first (or set the MTU on the Docker network instead).
3. **Single node only.** The `docker exec` targets `-control-plane` alone. `infra/kind/cluster-config.yaml` declares one node today, so this works — but loop over `kind get nodes --name $(KIND_CLUSTER)` and it keeps working the day someone adds a worker.

infra/kind/cluster-config.yaml
```diff
@@ -0,0 +1,15 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  # Default 65535 breaks TLS handshakes with external sites (GitHub, LinkedIn CDN)
```
gabriel-adopt: This comment describes a fix the file doesn't contain. It sits inside a `networking:` block with no MTU key — Kind has no MTU knob, which is *why* `kind-fix-mtu` shells out to `sed`. A reader lands here, sees the MTU rationale, and concludes the cluster config handles it.

Suggest rewording to say the MTU is patched post-create by `make kind-fix-mtu` because Kind exposes no setting for it, and pointing at that target.

Also worth noting: `podSubnet: 10.244.0.0/16` and `serviceSubnet: 10.96.0.0/16` are Kind's own defaults, so those two lines are no-ops. Harmless, but pinning defaults explicitly is a reasonable thing to *say* you're doing rather than have a reader assume they're load-bearing.

infra/kind/cluster-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+  podSubnet: "10.244.0.0/16"
+  serviceSubnet: "10.96.0.0/16"
+nodes:
+  - role: control-plane
+    kubeadmConfigPatches:
+      - |
+        kind: ClusterConfiguration
+        apiServer:
+          extraArgs:
+            "service-node-port-range": "1-65535"
```
gabriel-adopt: What needed this? `nodePort` appears nowhere in `charts/` or `infra/`, and local access goes through `kubectl port-forward` (`Makefile:217`).

Opening `1-65535` lets a Service claim privileged ports (<1024) and the entire ephemeral range, which invites collisions with kubelet's own ports and with host services — an odd trade for a range nothing currently uses. I'd drop it, or narrow it and note what required the widening.

apps/api/src/modules/auth/bootstrap.service.ts
```diff
@@ -26,7 +27,9 @@ export class BootstrapService implements OnModuleInit {
   ) {}
 
   async onModuleInit() {
-    const tenantCount = await this.tenantRepo.count();
+    const tenantCount = await this.tenantRepo.count({
```
gabriel-adopt, line 32: The fix is right, but the class docstring above is now false: line 15 still reads *"Idempotent: skips if any tenant exists"* when the rule is now any **non-system** tenant. That comment is what the next reader trusts over the query — please update it in the same commit.

Two notes beyond this PR's scope, for the record:

- The exclusion is hard-coded to one system tenant ID. Seed a second system tenant someday and bootstrap silently reverts to being skipped on fresh installs — the same bug, new phantom. An `is_system` flag on the tenant row is the durable fix.
- Bootstrap now re-attempts on **every** boot until a real tenant exists, where before the system tenant permanently suppressed it. Not reachable today (the recording pool creates apps and sessions under the system tenant, not users), but if a user ever existed there with `ADMIN_BOOTSTRAP_EMAIL`, `userRepo.save` would turn this into an unhandled throw in `onModuleInit` — i.e. an API crash-loop rather than a skipped bootstrap.

apps/controller/src/reconcile.service.ts
```diff
@@ -218,12 +220,13 @@ export class ReconcileService implements OnModuleInit, OnModuleDestroy {
 
     // Keep egress allowlist synced for all currently active runtime sessions.
     const { extraAllowlist, allowAll } = this.resolveEgressOptions(app);
+    const effectiveAllowAll = this.disableNetworkPolicy || allowAll;
```
gabriel-adopt, line 223: This is the second half of the flag and it has no test. The two new specs both drive `provisionSessionRuntime`, so `effectiveAllowAll` on the reconcile path — the one that keeps *already-active* sessions synced at :229 — is uncovered. Worth one more case, since this is the path that runs continuously rather than once at creation.

One asymmetry while you're here: `needsWorkerHealth` at :385 still keys off the raw `allowAll` (i.e. `browser_policy.recording_mode`) rather than `effectiveAllowAll`. That's *correct* as far as I can tell — with the flag on there's no NetworkPolicy, so 8091 ingress isn't gated and the Service isn't needed — but the two names now differ by one word and mean different things three lines apart. A short comment saying why the raw value is deliberate would save the next reader the trip through `pod-manager`.

apps/controller/src/reconcile.service.spec.ts
```diff
@@ -251,6 +251,115 @@ describe('ReconcileService restart_requested', () => {
...
+    return {
+      id: 'sess-np',
+      tenant_id: 'tenant-np',
+      app_id: 'app-np',
+      pod_name: null,
+      state: 'STARTING',
+      state_version: 1,
+      retry_count: 0,
+      owner_user_id: null,
+      residential_proxy_override: null,
```
gabriel-adopt: This fixture field doesn't exist on the entity, so the assertion passes for the wrong reason. `resolveResidential` reads `session.residential_proxy_enabled ?? app.residential_proxy_enabled` (`reconcile.service.ts:301`) — `residential_proxy_override` is never consulted, and the expected `false` in the assertion comes from `undefined ?? false` collapsing correctly rather than from anything the fixture set.

Rename it to `residential_proxy_enabled: null` so the fixture documents the real contract and the test would actually fail if that precedence changed.

(The `process.env` snapshot-and-reassign pattern and the inline `podManager` mocks match the file's existing style at :49-62 and :207, so no change wanted there — flagging only so it's clear that was checked and deliberate.)

charts/browser-hitl/values.yaml
```diff
@@ -484,6 +484,9 @@ config:
   reconcileBatchSize: "50"
   extractTabTimeoutMs: "15000"
   extractTabPollIntervalMs: "3000"
+  # Local dev: skip K8s NetworkPolicy creation and force egress allow_all.
+  # Leave "false" in staging/production — NPs are essential for isolation.
+  disableNetworkPolicy: "false"
```
gabriel-adopt, line 497: Two "network policy" switches with opposite polarity now live in this file: `networkPolicies.enabled` at :391 (chart-level inter-service NPs) and `config.disableNetworkPolicy` here (per-session `np-{sessionId}` NPs). Similar names, inverted senses, different scopes — that invites exactly the misconfiguration this PR is trying to prevent.

Prefer a positive `config.sessionNetworkPolicies: "true"` (default on), or at minimum a comment on each pointing at the other and stating which layer it governs.

**Do not derive one from the other**, in case that comes up in review: `networkPolicies.enabled` defaults to `false` here ("Enable in production"), so `DISABLE_NETWORK_POLICY: {{ not .Values.networkPolicies.enabled }}` would silently switch off per-session NPs in environments that currently get them. The two flags genuinely need to stay independent — which is all the more reason for the naming to make that obvious.

gabriel-adopt (commented):
## Review

Six independent local-dev fixes in one branch. I read every changed file plus the surrounding code on this head, and cross-checked the claims rather than the description.

### What holds up

- **`DISABLE_NETWORK_POLICY` blast radius is as advertised.** The key reaches only the shared ConfigMap (`configmap.yaml:103`), consumed by the controller via `envFrom` (`controller-deployment.yaml:37-39`), defaults to `"false"` in `values.yaml:489`, and is `"true"` only in `values-local.yaml`. Nothing in `.github/workflows/deploy-*.yaml` or `infra/tfy/deploy.yaml` overrides it.
- **The disable branch is semantically correct, not just a skip.** `createNetworkPolicy` already calls `syncEgressAllowlist` internally (`pod-manager.service.ts:218`), so replicating that call preserves the allowlist push. Teardown is safe too: `deletePolicy` swallows NotFound (`pod-manager.service.ts:872`), so `clearEgressAllowlist` still runs for sessions that never had an NP — no leaked `allow_all` entries.
- **The `server.js` edit will actually reach running pods** — `egress-proxy-deployment.yaml:31` has the `checksum/config` annotation, so the subPath-ConfigMap staleness trap doesn't apply.
- **Every Swagger claim checks out against `dsl.validator.ts`**: `interval_seconds >= 60` (:222), `health_checks` non-empty (:234), `actions` array-but-may-be-empty (:226), `policy` set + `quorum_n >= 1` (:263-268), the 5 `artifact_types` (:323), `AES-256-GCM` (:337), `ttl_seconds >= 300` (:341), `credential_ref` prefixes (:189), "at least one goto" (:195), and the 16-action list matches `VALID_ACTIONS` verbatim. `{ channels: [] }` is valid — validation is gated on `length > 0` (:364).

### Verdict

**Approve with changes.** One gating item: the CloakBrowser bump is the only change here that reaches production, its failure mode is silent, and CI structurally cannot catch it (no Docker build job) — see the comment on `apps/worker/package.json`. The rest are inline, ranked roughly by cost of getting them wrong.

### Process notes

- **`kind-fix-dns` is absent from the PR description** — it arrived in `d735bbf`, and the Files-changed table still lists the Makefile row as `kind-fix-mtu` only. Hardcoded public resolvers are something a reviewer should be shown, not find. Please refresh the body.
- **No docs update.** `CLAUDE.md` is the canonical local-dev guide (and this retro's audience), and devs with an *existing* Kind cluster must run `make kind-fix-mtu` and `make kind-fix-dns` by hand. That's the retro's core knowledge — it belongs next to the `make k8s-port-forward` line, not only in Makefile help text.
- Base `main` matches `CONTRIBUTING.md:116`, though `CLAUDE.md:288` claims the main branch is `dev`. Worth reconciling those docs separately.
- Branch is 10 commits behind `origin/main`; none touch these files, so no conflict — just rebase before merge.
- Also confirmed: all 7 CI checks green, and Copilot's 4 earlier comments were genuinely addressed in `b8e2cc4` rather than papered over.

### Nits (not worth blocking)

- `key_ref format: k8s:secret/{name}` is documented but not enforced anywhere in `dsl.validator.ts` — fine as convention, just don't let it read as validation.
- `kind-fix-mtu` uses `--timeout=30s`, `kind-fix-dns` uses `60s`; no reason for the difference.
- `kubectl get cm coredns -o yaml | kubectl apply -f -` will warn about the missing `last-applied-configuration` annotation on a kubeadm-created object. Cosmetic.
- The direct CONNECT path has no `setTimeout`, unlike the residential path's `UPSTREAM_CONNECT_TIMEOUT_MS` — a black-holed SYN still hangs ~2min with no response, adjacent to the confusion this PR fixes. Out of scope; worth a follow-up.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

navjot-adopt:
Thanks for the thorough pass, Gabriel — all actionable items addressed in a9ae1a1.

**Gating item — CloakBrowser 0.5.2 proof of change:**
- `make docker-build-worker` → exit 0. The Dockerfile step `import { ensureBinary } from 'cloakbrowser'; await ensureBinary()` builds clean on 0.5.2, so the named export and cache-provisioning path still exist.
- Verified inside the built image: `ensureBinary: function | launch: function`, `CLOAKBROWSER_CACHE_DIR=/opt/cloakbrowser` honored, and `/opt/cloakbrowser` holds the chromium binaries (`chromium-146.0.7680.177.*`) where the runtime looks.
- A live worker session logged `Browser launched via CloakBrowser (stealth mode)` (no fallback to stock Playwright).
- Added all three to the PR test plan.

**Safety — the big one:** you're right that `kind-fix-dns`/`kind-fix-mtu` could hit a remote cluster. Both now pin `--context kind-$(KIND_CLUSTER)` on every kubectl call and share a `kind-guard` that refuses to run if that context is absent. Verified: guard rejects a bogus cluster name; targets run clean against `kind-tabby-dev`.

**Verify-then-report:** `kind-fix-dns` now greps the rendered Corefile for `8.8.8.8` and fails loudly if the sed was a no-op. `kind-fix-mtu` restarts kindnet **first** (your ordering point — kindnet rewrites the conflist on startup), then patches every node from `kind get nodes` and greps each for `"mtu": 1500`, failing if it didn't land.

**Correctness:**
- `residential_proxy_override` fixture → `residential_proxy_enabled: null` (documents the real precedence; would now fail if it changed).
- Added a reconcile-path test covering `effectiveAllowAll` at :229 (the continuous path), and a comment on why `needsWorkerHealth` deliberately keys off raw `allowAll`.

**Docs/clarity:** bootstrap docstring corrected ("any *non-system* tenant"); `cluster-config.yaml` MTU comment fixed + dropped the unused `service-node-port-range 1-65535` + noted the subnets are Kind defaults; cross-referencing comments on both network-policy flags in `values.yaml`; egress-proxy now `log('direct CONNECT failed', …)` before the 502 with a JSON body matching `fail()`; PR description refreshed with `kind-fix-dns`; CLAUDE.md documents both targets + MTU/CoreDNS gotchas (#24, #25).

**Out-of-scope follow-ups filed:** Sentry-capture the CloakBrowser fallback (#148), `is_system` tenant flag (#149), direct-CONNECT `setTimeout` (#150), pnpm pin / lockfile churn + unmet socks-proxy-agent peer (#151).

Nits (`key_ref` is convention not validation, timeout 30s/60s, `last-applied-configuration` warning) noted — left as-is unless you'd like them tightened.

apps/worker/package.json
navjot-adopt: Proof added in a9ae1a1 + PR test plan: `make docker-build-worker` exits 0 (Dockerfile runs `ensureBinary()` on 0.5.2); inside the built image `ensureBinary: function | launch: function`, `CLOAKBROWSER_CACHE_DIR=/opt/cloakbrowser` honored with the chromium binaries present; and a live worker session logged `Browser launched via CloakBrowser (stealth mode)`. Filed #148 to turn the silent `console.warn` fallback into a Sentry capture.

charts/browser-hitl/files/egress-proxy/server.js
navjot-adopt: Fixed in a9ae1a1 — the handler now `log('direct CONNECT failed', { session_id, target, reason })` before responding, and the 502 uses a JSON body (`{error, reason}`) with `Content-Type`/`Content-Length` matching residential `fail()`. The missing direct-CONNECT `setTimeout` is a real gap too — filed #150.

Makefile
navjot-adopt: Fixed in a9ae1a1 — both `kind-fix-*` targets now pin `--context kind-$(KIND_CLUSTER)` on every kubectl call and share a `kind-guard` that refuses to run if that context is absent, so they can never touch a remote cluster. Verified: the guard rejects a bogus cluster name; targets run clean against `kind-tabby-dev`.

Makefile
navjot-adopt: Fixed in a9ae1a1 — after apply, `kind-fix-dns` greps the rendered Corefile for `8.8.8.8` and fails loudly if the sed was a no-op, instead of reporting success unconditionally.

Makefile
navjot-adopt: Fixed in a9ae1a1 — `kind-fix-mtu` now restarts kindnet FIRST (your ordering point — kindnet rewrites the conflist on startup), then patches every node from `kind get nodes` and greps each for `\"mtu\": 1500`, failing if the substitution did not land. Covers the silent-no-op, ordering, and single-node points.

infra/kind/cluster-config.yaml
navjot-adopt: Fixed in a9ae1a1 — reworded to say Kind exposes no MTU setting so it is patched post-create by `make kind-fix-mtu`, and noted the two subnets are Kind defaults pinned only for explicitness.

infra/kind/cluster-config.yaml
navjot-adopt: Fixed in a9ae1a1 — dropped the `service-node-port-range: 1-65535` kubeadm patch entirely. Nothing uses nodePort (local access is `kubectl port-forward`), so the widened range is gone.

apps/api/src/modules/auth/bootstrap.service.ts
navjot-adopt: Fixed in a9ae1a1 — docstring now reads \"skips if any *non-system* tenant exists\" and explains the RECORDING_POOL exclusion. Filed #149 for the durable `is_system` flag (and the crash-loop note).

apps/controller/src/reconcile.service.ts
navjot-adopt: Fixed in a9ae1a1 — added a reconcile-path test that drives `reconcileApp` and asserts `syncEgressAllowlist` is called with `allowAll=true` for an already-active session (the :229 path). Also added a comment explaining why `needsWorkerHealth` deliberately keys off raw `allowAll`, not `effectiveAllowAll`.

apps/controller/src/reconcile.service.spec.ts
navjot-adopt: Fixed in a9ae1a1 — renamed the fixture field to `residential_proxy_enabled: null` with a comment documenting the `session ?? app` precedence, so the test would now fail if that changed.

charts/browser-hitl/values.yaml
navjot-adopt: Fixed in a9ae1a1 — added cross-referencing comments on both flags (`networkPolicies.enabled` = chart-level inter-service NPs; `config.disableNetworkPolicy` = per-session NPs), each pointing at the other with scope + polarity. Kept independent per your note — not derived.

rahulbh1510 (approved):

PR #157, opened 2026-08-04
feat: browser-driven skill reliability and bank-portal hardening

navjot-adopt:
## Summary

Everything needed to make **browser-driven skills** (`call_web_browser` → Tabby `/execute/browser`) work reliably against refresh-sensitive, bot-protected bank portals (HSBCnet, ICICI, HDFC Life), plus the supporting worker/controller/api/egress reliability fixes surfaced along the way.

> Squash-merged, so per-commit history collapses — grouped summary below.

### Browser-driven skill support
- **`activity` keepalive** — trusted mouse/scroll idle-reset that holds SPA/bank sessions **without a full-page reload** (a `goto` keepalive re-handshakes single-session portals → "already logged on"). Health-gated so it never nudges a login page. (`2a1e5df`, `1181de5`)
- **`click_by_text` robustness** — visible-first + substring matching for duplicate/near-miss labels; **DOM-dispatched click-through** when an overlay intercepts the click; **`get_page_summary` returns only visible elements** so hidden/stale modals stop misleading the model. (`767f6b1`, `3369e44`)
- **Download capture** — `download-capture` + `list_downloads`/`get_download` on `/execute/browser` (base64), gated on `browser_policy.downloads`. (`71f8b13`)

### Anti-bot / egress hardening (global CHROMIUM_FLAGS)
- `--disable-quic` so bank traffic can't bypass the HTTP-CONNECT egress proxy over HTTP/3. (`c4d21ae`)
- Drop `--enable-automation`, add Windows fonts, SwiftShader software WebGL (present a real GPU renderer). (`20022dd`, `aee9818`)
- Egress-proxy Happy-Eyeballs attempt budget raised for distant hosts. (`9098c7e`)

### Health-check correctness
- `url_check` catches client-side SPA session expiry; `dom_check` tests DOM **presence** (not CSS visibility), handles multi-match, and `exists:false` can pass; health-check config fields declared. (`1563c1a`, `fa9a030`, `d22c267`, `91ec3b2`, `bbdc189`)

### Controller / API reliability
- FAILED session no longer blocks its own replacement (dead-zone). (`b9ace4b`)
- `session-status` never returns a TERMINATED session as current. (`428a764`)
- App-template updates propagate from a fresh read. (`0466ee0`)
- MinIO orphan sweep no longer deletes recording bundles. (`6560864`)

### Viewer
- Human health label instead of the raw `AUTH_FAIL` enum; noVNC vendored into the image; stop claiming "Signed in" while recording. (`ec8174d`, `1a23508`)

### Cross-origin execute
- Route cross-origin fetches off-page so CORS can't block them (`/execute/fetch`). (`b67341e`)

### Local dev tooling (no stage/prod effect)
- Makefile: pin kube-context on kind targets, fix the MTU patch race, pin `--context` on port-forward; optional gitignored local-secrets override. (`a6ad3d4`, `14af184`, `045dcef`)
- Local egress allowlist entries for the test bank portals in `values-local.yaml`. (`ae38444`, `6e15a1c`)

## Non-browser flow impact
Browser-driven-specific changes (click_by_text, visible summary, download capture, activity keepalive) are **additive / opt-in / browser-scoped** — HAR-replay `call_web_api`, login-DSL and other non-browser flows are unaffected. Three globally-scoped changes touch shared paths and warrant a non-browser regression check before stage: **CHROMIUM_FLAGS** (every browser session), **dom_check semantics** (presence vs visibility — used by all health checks), and **cross-origin fetch off-page** (`/execute/fetch`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,6 @@ docs/zero-to-hero-test-workflow.md
 docs/zero-to-hero-tabby-platform-setup.md
 docs/adopt-org-templates-backup.json
 docs/tabby-platform-setup-consolidation.md
+
+# Local-only Kind secrets override (never commit — real credentials)
+charts/browser-hitl/values-local.secret.yaml
--- a/Makefile
+++ b/Makefile
@@ -260,19 +260,45 @@ kind-guard:
 
 .PHONY: kind-fix-mtu
 kind-fix-mtu: kind-guard ## Fix Kind pod MTU from 65535 to 1500 (prevents TLS failures to external sites)
-	@echo "Patching kindnet MTU to 1500 on all nodes of '$(KIND_CLUSTER)'..."
-	@# kindnet rewrites its conflist on startup, so restart the daemonset FIRST,
-	@# then patch every node's conflist, so the patch is what survives.
+	@echo "Patching kindnet pod MTU to 1500 on all nodes of '$(KIND_CLUSTER)'..."
+	@# Patch ONLY the CNI conflist (pod veth MTU). Do NOT touch the node's own eth0:
+	@# it must keep matching the Docker bridge (`docker network inspect kind` ->
+	@# com.docker.network.driver.mtu, 65535 on Docker Desktop). Lowering the node
+	@# interface while the bridge stays 65535 makes the bridge emit frames the node
+	@# drops, which wedges kubectl port-forward with
+	@#   'error creating error stream ...: Timeout occurred'
+	@# and takes the API/VNC tunnels down with it.
+	@#
+	@# kindnet rewrites the conflist on startup, and does so a moment AFTER the
+	@# daemonset reports Ready — so a patch applied immediately gets clobbered.
+	@# Restart first, then patch, then RE-verify once kindnet has settled, retrying
+	@# before giving up. The previous version verified instantly and printed success
+	@# against a file that was about to be overwritten.
 	kubectl --context $(KIND_CONTEXT) rollout restart daemonset/kindnet -n kube-system
-	kubectl --context $(KIND_CONTEXT) rollout status daemonset/kindnet -n kube-system --timeout=30s
-	@for node in $$(kind get nodes --name $(KIND_CLUSTER)); do \
-		docker exec $$node sh -c '\
-			CNI=/etc/cni/net.d/10-kindnet.conflist; \
-			[ -s "$$CNI" ] && sed -i "s/\"mtu\": *[0-9]*/\"mtu\": 1500/g" "$$CNI"; \
-			grep -q "\"mtu\": 1500" "$$CNI"' \
-			|| { echo "MTU patch did not land on $$node (no \"mtu\" key? kindnet variant changed)"; exit 1; }; \
-	done
-	@echo "MTU set to 1500 on all nodes (verified). Restart pods to pick it up."
+	kubectl --context $(KIND_CONTEXT) rollout status daemonset/kindnet -n kube-system --timeout=60s
+	@ok=0; \
+	for attempt in 1 2 3 4 5; do \
+		for node in $$(kind get nodes --name $(KIND_CLUSTER)); do \
+			docker exec $$node sh -c '\
+				CNI=/etc/cni/net.d/10-kindnet.conflist; \
+				[ -s "$$CNI" ] && sed -i "s/\"mtu\": *[0-9]*/\"mtu\": 1500/g" "$$CNI"; \
+				grep -q "\"mtu\": 1500" "$$CNI"' \
+				|| { echo "MTU patch did not land on $$node (no \"mtu\" key? kindnet variant changed)"; exit 1; }; \
+		done; \
+		sleep 20; \
+		stable=1; \
+		for node in $$(kind get nodes --name $(KIND_CLUSTER)); do \
+			docker exec $$node sh -c 'grep -q "\"mtu\": 1500" /etc/cni/net.d/10-kindnet.conflist' || stable=0; \
+		done; \
+		if [ "$$stable" = "1" ]; then \
+			echo "Pod MTU 1500 verified stable on all nodes (attempt $$attempt)."; ok=1; break; \
+		fi; \
+		echo "kindnet reverted the patch (attempt $$attempt) — retrying..."; \
+	done; \
+	if [ "$$ok" != "1" ]; then echo "MTU patch keeps being reverted by kindnet; aborting."; exit 1; fi
+	@echo "NOTE: kindnet rewrites this on every restart — re-run after one."
+	@echo "Existing pods keep their old MTU — recreate them to pick this up:"
+	@echo "  kubectl --context $(KIND_CONTEXT) rollout restart deploy -n $(HELM_NAMESPACE)"
 
 .PHONY: kind-fix-dns
 kind-fix-dns: kind-guard ## Point CoreDNS at public resolvers (fixes AAAA SERVFAIL from Docker Desktop embedded DNS)
@@ -306,10 +332,10 @@ kind-load-images: ## Load all Docker images into the Kind cluster
 	@echo "Images loaded into Kind cluster '$(KIND_CLUSTER)'"
 
 .PHONY: kind-deploy
-kind-deploy: ## Full local deploy: load images + helm install with local values
+kind-deploy: kind-guard ## Full local deploy: load images + helm install with local values
 	$(MAKE) kind-load-images
-	kubectl create namespace $(HELM_NAMESPACE) --dry-run=client -o yaml | kubectl apply -f -
-	helm upgrade --install $(HELM_RELEASE) charts/browser-hitl/ \
+	kubectl --context $(KIND_CONTEXT) create namespace $(HELM_NAMESPACE) --dry-run=client -o yaml | kubectl --context $(KIND_CONTEXT) apply -f -
+	helm upgrade --install $(HELM_RELEASE) charts/browser-hitl/ --kube-context $(KIND_CONTEXT) \
 		-f charts/browser-hitl/values-local.yaml \
 		--namespace $(HELM_NAMESPACE) \
 		--wait --timeout 5m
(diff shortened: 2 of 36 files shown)


navjot-adopt:
## Deep review — findings before stage

Full-diff review vs `dev`. Builds and tests pass on the branch (worker 151, api 607, controller 77, shared 103). Items below are code-level defects, grouped by severity. I have **not** changed anything yet.

### 🔴 Blocking

**1. `apps/controller/src/reconcile.service.ts:226` — FAILED sessions no longer count toward capacity, but their pods are never deleted → pod leak + double-provisioning.**
The FAILED transition only writes DB state; teardown happens solely in the FAILED-cleanup pass at `:553`, which is gated on `globalIdleMs > 0`. **Verified:** `IDLE_SHUTDOWN_SECONDS` defaults to `"0"` (`values.yaml:450`, `configmap.yaml:31`) and `reconcile.service.ts:552` is `if (globalIdleMs > 0)`, so that pass never runs by default. `reconcileRuntimeDrift` only reaps pods whose session is missing/TERMINATED, so FAILED isn't covered either.
*Scenario:* app `desired=1`, session FAILs → its worker pod runs forever → next tick sees `actual=0` and creates another. Grows until the circuit breaker caps it (~5 zombie pods/app/15min). Pre-change, the FAILED session counted as capacity so only 1 pod existed.
*Fix:* tear the pod down on the transition into FAILED, or run the FAILED reap unconditionally with its own TTL.

**2. `apps/worker/src/health-predicate-runner.ts:203` — `dom_check exists:true` changed from CSS-visible to DOM-attached; stored tenant configs can now report HEALTHY while logged out.**
Old: `waitFor()` (default `visible`) + `isVisible()`. New: `waitFor({state:'attached'})`. Any config whose logged-in marker stays in the DOM after a client-side logout (SPA route swap that hides rather than unmounts, `display:none` shells, pre-rendered nav) now PASSes on the login page → session stays HEALTHY, never goes to LOGIN_NEEDED, no HITL prompt. In-repo defaults (`selector:'body'`) are unaffected; **DB-stored tenant configs are the risk**.
*Fix:* make it an explicit per-check mode (`match: 'attached' | 'visible'`, defaulting to `visible` for existing rows) rather than redefining `exists` globally.

**3. `packages/shared/src/constants.ts:35` — dropping `--enable-automation` doesn't remove it; Playwright re-injects it.**
`playwright-core` appends `--enable-automation` unless `ignoreDefaultArgs: ['--enable-automation']` is set, and `main.ts:132` sets none. On the stock-Playwright fallback path (taken whenever CloakBrowser fails to load — which only `console.warn`s) `navigator.webdriver` stays `true`. The comment claims the opposite, so the mitigation reads as done when it isn't.
*Fix:* add `ignoreDefaultArgs: ['--enable-automation']` (and check what CloakBrowser does with it).

### 🟡 Non-blocking

4. `execute-browser-handler.ts:248` — `click_by_text` default for `exact` silently flipped `true` → `false`. Already-compiled skills that omit `exact` change behavior (e.g. "Log out" now also matches "Log out of all devices"; `nth(0)` picks DOM order). Suggest: keep exact-first, fall back to substring only on 0 matches.
5. `keepalive-runner.ts:110` — the activity gate fires on `!== 'PASS'`, which includes `TRANSIENT_FAIL`. A 5xx/timeout/egress blip disables the idle nudge and can let the portal's idle timer expire — turning a recoverable blip into a real expiry. The bot-scoring rationale only applies to `AUTH_FAIL`; suggest gating on that.
6. `download-capture.ts:94` — captured files are never unlinked (only the index entry is spliced). A long-lived worker with downloads enabled accumulates every export on ephemeral disk.
7. `download-capture.ts:83` — `id: dl-${Date.now()}-${records.length}` reads length **before** push; once the index caps at 20 that component is constant, so same-millisecond downloads can collide.
8. `streaming.controller.ts:2110` — the new containment `throw` is inside the `try`, so the bare `catch` swallows it and then fetches the same unvalidated path from the CDN. Not exploitable (the upstream `normalizeNoVncAssetPath` is the real control) but the guard denies nothing; also means a broken vendor dir silently falls back to the public CDN with no log.
9. `execute-handler.ts:80` — `new URL('about:blank').origin === 'null'`, so while the page is on `about:blank` every `/execute/fetch` routes off-page; off-page traffic never fires `page.on('request')`, so `har_start`→fetch→`har_stop` records **zero** entries for cross-origin calls.
10. `reconcile.service.ts:575` — the per-template `idle_shutdown_seconds` lookup is effectively dead (`appById`/`templateById` are built only from HEALTHY sessions, and the branch only runs when `liveCount === 0`), so the global env value always wins.
11. `agent.service.ts:100` — `getSessionStatus` now 404s on FAILED too; combined with (1), a persistently-failing app reports "no live session" indefinitely instead of surfacing the failure.
12. `execute-browser-handler.ts:66` — `click_element` didn't get the same hardening as `click_by_text` (no visible-filter, no overlay `dispatchEvent` fallback), so the sticky-banner case still breaks the selector path.

### 🧪 Test gaps
- **`get_page_summary` visibility filtering has no test** — the `page.evaluate` body never executes in the suite, so `checkVisibility` / `aria-hidden` / filter-before-slice are unverified.
- `main.ts` `acceptDownloads` wiring + `enableDownloadCapture` vs cancel path: no test.
- The new noVNC path containment / CDN fallback: no behavioral test (the added tests string-match the `.ts` source).
- `execute-integration.spec.ts:376` mocks a resolved response, so it passes even if `failOnStatusCode:false` were deleted; no test asserts the options passed to `context.request.fetch`.
- `download-capture.spec.ts` writes real files into `os.tmpdir()` and never cleans up.

### On the three non-browser questions from the description
- **`--disable-quic`:** safe — forces HTTP/1.1+2 over TCP, which the CONNECT proxy already handles. Replay + login DSL unaffected.
- **SwiftShader/ANGLE:** spawns a real GPU process with software raster on every pod against the 1536Mi limit — **unmeasured**; worth a memory check before stage.
- **`dom_check`:** the `exists:false` direction is a strict improvement (it could never pass before). The `exists:true` direction is the regression in (2).
- **Cross-origin fetch:** same-origin replay is byte-identical (in-page path untouched, only gains an off-page retry). Two deltas: the `about:blank` case (9), and off-page responses expose the full header set incl. `set-cookie` where an in-page fetch only surfaced CORS-exposed headers.

navjot-adopt:
### Addressed in `c5aa848` — blocking 1 and 2

**1. FAILED-session pod leak.** `releaseSessionRuntime()` is split out of `terminateSession()` and now runs for any FAILED session still holding a pod, clearing `pod_name` so the reap is idempotent.

Deliberately **state-driven** (in `reconcileApp`, off the session's state) rather than hooked to the FAILED transition, so it self-heals no matter who marked the session failed — controller, worker, or API — and it runs every tick regardless of `IDLE_SHUTDOWN_SECONDS`. A reap failure is logged and retried next tick rather than blocking provisioning.

The FAILED session still doesn't hold capacity, so replacement provisioning is unchanged — one of the new tests asserts both in the same tick (pod deleted **and** a fresh session created).

**2. `dom_check` semantics.** Added an explicit `match: 'attached' | 'visible'`, defaulting to `attached`.

I kept `attached` as the default rather than reverting to `visible`: `runDomCheck`'s own evidence (a signed-in ICICI dashboard resolving its sidebar link to `hidden` 20× in a row, plus Salesforce Lightning and Workday) makes strict visibility the dominant false-AUTH_FAIL on this platform, so defaulting to it would trade one regression for a worse one. `match: 'visible'` is the documented opt-out for portals that hide rather than unmount their logged-in chrome on sign-out.

Also mirrored on the negative check: under `match: 'visible'`, "the logged-out marker is gone" now waits for Playwright's `hidden` (detached **or** invisible) — which is what the comment at the top of `runDomCheck` always claimed while the code used `detached` (the nit in the review). And `DomCheck` now declares the `timeout_ms` it was already reading.

**Verification:** worker 154 passed, controller 80 passed (6 new tests), full `pnpm run build` green across all 7 projects.

**Still open from the review:** non-blocking 4–12 and the test gaps — notably 4 (`click_by_text` default `exact` flipped `true`→`false`, which changes behavior for already-compiled skills) and 3 (`--enable-automation` is re-injected by Playwright unless `ignoreDefaultArgs` is set, so that mitigation is currently a no-op on the stock-Playwright path).

navjot-adopt:
### Non-blocking findings addressed in `49e1f90`

**4 + 12 (click precision).** `click_by_text` defaults to **exact-first, substring as a rescue** — that keeps already-compiled skills precise ("Log out" no longer also matches "Log out of all devices") while still rescuing the near-miss labels substring matching exists for ("Credit Cards" vs "Credit Card"). `click_element` now shares the same visible-first + overlay-dispatch path via `clickThroughOverlays()`; the sticky-banner interception was never text-specific.

**5 (keepalive gate).** Narrowed to `AUTH_FAIL` only. Gating on any non-PASS also caught `TRANSIENT_FAIL` — a 5xx, probe timeout or egress blip — where the session is still signed in, so the nudge stopped and the portal's own idle timer could expire: a recoverable blip became a real expiry, the exact outcome `activity` exists to prevent.

**6 + 7 (download hygiene).** Evicted downloads are unlinked (trimming the index left the bytes on disk forever), and ids use a monotonic counter (`records.length` is read before the push, so past the cap it was constant and same-millisecond downloads collided).

**8 (guard placement).** The noVNC containment check moved **above** the `try` — inside it the bare `catch` swallowed the rejection and then fetched the same unvalidated path from the CDN, so the guard re-routed rather than denied. The CDN fallback is now logged too; using it silently defeats the point of vendoring noVNC.

**9 (opaque origin).** `about:blank` serializes to the **string** `"null"`, so off-page routing was correct only by accident. Behaviour unchanged, but explicit — and the HAR consequence (off-page requests never fire `page.on('request')`) is documented at the site.

**10 (dead lookup).** Removed the per-template idle lookup in the FAILED-cleanup stand-down: `appById`/`templateById` are built only from apps of HEALTHY sessions and the branch only runs when there is none, so it always fell back to the global value.

**Deliberately not changed — 11** (`getSessionStatus` excluding FAILED). Returning a dead session as "current" is what made the harness mint a sign-in card pointing at a terminated session; with FAILED pods now released promptly, "none — provision fresh" is the right answer. A `usable: false` shape is an API change for its own PR.

**Verification:** worker 157, controller 80, api 607 — `pnpm run build` and `nx run-many -t test` green across all 7 projects.

rahulbh1510 (approved):

PR #159, opened 2026-08-07
fix(worker): don't inject keepalive input while an agent is driving the page

navjot-adopt:
## Problem

Keepalive exists to stop a portal's idle timer from expiring an **unattended** session. When the harness is executing browser commands, that premise is false twice over:

1. **The nudge is redundant** — the agent's own clicks and navigations already reset the portal's idle timer.
2. **It is destructive.** A synthetic mouse-move/scroll landing between the agent's locator resolution and its click misplaces the target and misclicks. A `goto` keepalive reloads the page out from under a half-finished flow. And robotic input *interleaved with real interaction* reads worse to reCAPTCHA v3 / fingerprint SDKs than either pattern alone.

## Fix

New `apps/worker/src/agent-activity.ts` tracks agent commands (`/execute/browser` and `/execute/fetch` report into it). The keepalive loop gates on **two distinct signals**:

### 1. Command in flight → skip the whole cycle, health included

Not just the actions. `runDomCheck` resolves a locator against the live page and returns **`AUTH_FAIL`** when it cannot find it (`health-predicate-runner.ts:214`). A health check evaluated mid-navigation therefore reports a *signed-in* session as signed **out**, which drives it to `LOGIN_NEEDED` and shows the human a sign-in card in the middle of a working task.

Capped at 3 consecutive skips so health cannot stall behind a continuously busy agent.

### 2. Agent active within the keepalive interval → drop the actions

If the agent clicked 20s ago on a 120s interval, the nudge is redundant.

## The judgement call worth reviewing

Read-only commands (`screenshot`, `get_page_summary`, `get_page_info`, `scroll_page`) count as **busy** but **not** as **activity**.

They make no HTTP request, so the portal's idle timer keeps running and the session genuinely still needs the nudge — this is gotcha #15 (*"`screenshot` keepalive does not keep sessions alive"*) applied in the other direction. Treating every command as activity would mean an agent sitting there screenshotting silently suppresses keepalive until the session expires.

They still make the agent *busy*, though: a scroll injected during a `screenshot` or `get_page_summary` corrupts what those return.

## Also

On-demand artifact extraction is held off while a command is in flight — it navigates the page when `export_policy.extract_urls` is set, which would yank the agent off its task as hard as a `goto`. The request stays in Redis and is picked up by the next 2s poll.

## Blast radius

**Sessions with no agent traffic are unaffected.** `msSinceAgentActivity()` returns `Infinity` when the agent has never touched the session, so every gate is a no-op and behaviour is byte-for-byte what it was. This matters because Tabby is a general platform, not harness-only — plain credential-vending sessions are the common case.

The existing `AUTH_FAIL` gate on the `activity` nudge (don't feed robotic input to a login page being scored by reCAPTCHA) is untouched and still applies.

## Testing

- 5 new cases in `keepalive-runner.spec.ts` covering: suppression inside the interval, resumption outside it, read-only commands not counting as activity, whole-cycle skip while in flight, and health running anyway after the skip cap.
- Full worker suite green: **162 tests, 13 suites**.
- `pnpm run build` + `pnpm -r run lint` clean.

### DIFF
--- a/apps/worker/src/agent-activity.ts
+++ b/apps/worker/src/agent-activity.ts
@@ -1,0 +1,88 @@
+/**
+ * Tracks whether an agent (the harness, via /execute/browser and /execute/fetch)
+ * is currently driving this worker's page, so the keepalive loop can stay out of
+ * its way.
+ *
+ * Keepalive exists to stop a portal's idle timer from expiring an unattended
+ * session. When an agent is actively working the page, that premise is false
+ * twice over:
+ *
+ *  1. The agent's own clicks/navigations already reset the portal's idle timer,
+ *     so the keepalive action is redundant.
+ *  2. Worse, it is destructive. A synthetic mouse-move/scroll landing between an
+ *     agent's locator resolution and its click moves the target and misclicks; a
+ *     'goto' keepalive reloads the page out from under a half-finished flow. The
+ *     robotic input is also exactly what reCAPTCHA v3 / fingerprint SDKs score,
+ *     and it is far more suspicious interleaved with real interaction than alone.
+ *
+ * Two distinct signals, because they answer different questions:
+ *
+ *  - `isAgentBusy()` — is a command in flight right now? Every command counts,
+ *    including read-only ones: a scroll injected during `screenshot` or
+ *    `get_page_summary` corrupts what those return.
+ *  - `msSinceAgentActivity()` — how long since the agent did something that
+ *    actually reset the portal's server-side idle timer? Only server-touching
+ *    commands count here. Per the platform gotcha that `screenshot` is not a
+ *    keepalive (it captures pixels, it makes no request), a session that has
+ *    only been screenshotted is still going stale and still needs the nudge.
+ *
+ * Module-level state is correct here: one worker pod owns exactly one page.
+ */
+
+/** Commands that hit the origin server, and so reset its idle timer. */
+const IDLE_RESETTING_COMMANDS = new Set([
+  'navigate',
+  'click_element',
+  'click_by_text',
+  'click_at',
+  'type_text',
+  'type_into_label',
+  'press_key',
+]);
+
+let inFlight = 0;
+let lastActivityAt = 0;
+
+/** True if the given browser command resets the origin's server-side idle timer. */
+export function isIdleResettingCommand(command: string): boolean {
+  return IDLE_RESETTING_COMMANDS.has(command);
+}
+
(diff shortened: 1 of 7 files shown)


navjot-adopt:
## Follow-up: fixed the general form of the same bug (`e30ee8e`)

The first commit only *dodged* the `dom_check` → false `AUTH_FAIL` race, and only for the case where **the agent** is the one navigating. The underlying defect was still there: `runDomCheck` mapped **every** failure to `AUTH_FAIL`, conflating a genuine "the logged-in marker is gone" with three classes that carry no auth signal whatsoever.

| Failure | Was | Now |
|---|---|---|
| Page navigated mid-locator (`Execution context was destroyed`) — any timed redirect, every SPA route change | `AUTH_FAIL` | `TRANSIENT_FAIL` |
| Page/context/browser closed — pod teardown | `AUTH_FAIL` | `TRANSIENT_FAIL` |
| Malformed selector in the app's `health_checks` | `AUTH_FAIL` | `TRANSIENT_FAIL` |
| Plain Playwright timeout — the marker really is gone | `AUTH_FAIL` | `AUTH_FAIL` (unchanged) |

The selector row is the one worth dwelling on: **a single typo in an app's `health_checks` turned every session for that app permanently "signed out"** — silent, total, and indistinguishable from a real auth failure to everything downstream. Note gotcha #7 in `CLAUDE.md` exists precisely because unquoted attribute selectors are an easy mistake to make here.

### Why this direction

`AUTH_FAIL` is not a neutral verdict. It drives the session to `LOGIN_NEEDED`, raises HITL, and puts a sign-in card in front of a human. Claiming it on no evidence is strictly worse than admitting the check did not run — which is what `TRANSIENT_FAIL` already means, and which the caller already handles.

### Conservative by construction

`classifyDomCheckError` returns `null` for anything it cannot *positively* identify, so an unrecognised error keeps the previous `AUTH_FAIL` behaviour. This only carves out the classes we can name; it does not flip the default. A plain timeout — the overwhelmingly common case, and the one the whole check exists to detect — is untouched.

6 new tests, including one asserting the timeout case still returns `AUTH_FAIL` so the carve-outs cannot swallow the genuine verdict. Worker suite: 168 tests, 13 suites, green.

navjot-adopt:
Superseded by #160, which branches fresh from `dev` (same two commits, cherry-picked clean and verified byte-identical). This branch carried a stale name from the already-merged #157.

PR #163, opened 2026-08-07
feat(worker): capture what identifies a control, and the primitives to replay it

navjot-adopt:
The recorder side of the browser-skill recording redesign. Rebased onto current `dev`; supersedes the single-commit version this PR opened with.

## Why

Browser-driven skills (ICICI, HSBCnet) needed several rounds of production iteration each, while HAR-replay skills come out of the same recorder in good shape. The difference is structural: a replay skill's contract is the **network**, which the recorder captures losslessly and which needs no interpretation. A browser skill's contract is the **DOM and the route**, and the recorder captured an impression of that — one guessed selector per click, no page state, no outcome.

Worse, it recorded a **decision** rather than **evidence**. `buildRichSelector` picked one selector in-page via a fixed ladder (last resort: the bare tag name) with no idea whether it matched one node or forty. The page is gone when the recording ends, so no later compiler improvement could revisit it.

## What changed (workflow recordings only)

**Locator candidates with match counts.** Each interaction now carries several ways to address the element — test-id, id, name, aria-label, role+accessible-name, label, text, css path — each with **how many nodes it matched at that instant**. `match_count > 1` makes ambiguity a compile-time fact instead of a production misclick. (`a.mb-0` matching forty elements on ICICI's dashboard is the canonical case.)

**Element evidence.** Role, accessible name, visibility, bounding box, and whether something was painted over its centre. Occlusion is the overlay-intercepted-click problem answered at the only moment it is knowable.

**Two fidelity fixes.** Evidence comes from the **actionable ancestor**, not the legacy "nearest node with an id or any class" — which on a modern page is the `<span>` inside the button. And the element is resolved through `composedPath()`, so anything inside a web component records as itself rather than the shadow host.

**Interaction outcomes.** Did it navigate and where, how many requests fired, when they settled, did a download start. The compiler previously inferred causality from bare timestamps. The window truncates at the next interaction, and attribution keys off `event_time` — a field filled and submitted inside the 500ms debounce flushes *after* the click, so `timestamp` would hand the click's navigation to the fill.

**Downloads and popups.** Neither was captured at all. A `blob:` download never touches the network so HAR cannot see it, and every listener was bound to a single `Page` — so a statement opening in a new tab made the human's clicks there invisible and the compiled skill stopped at the click that opened it.

## Two bugs found along the way

**Warm-pool sessions ignored the requested mode.** A pooled spare boots from the shared pool app, hardcoded `recording_mode: 'login'`, and bind never told it otherwise. noui has regression cover for the visible half of this and works around it by classifying bundles from content. That workaround cannot recover the invisible half: on the fast path — most sessions — every workflow-only capture stayed switched off. Bind now carries the mode, and the injected recorder gained an upgrade path (its guard was a plain "already installed, do nothing", which would have pinned a bound spare to whatever it booted as).

**The HAR reduction was gated on the wrong axis.** `recording_mode` is the authoring *phase*; `browser_driven` is the skill *kind*. Gating on `workflow` gutted the HAR for every **replay** skill built from a post-login recording — `har_to_tools.py` reads exactly the fields the reduction empties — silently, with nothing erroring. It is now a separate flag defaulting to **false**, and additionally requires workflow mode so a login recording's HAR (the contract for building the profile) can never be reduced whatever the caller asks. A test asserting that caught it before it shipped.

## HAR

For browser-driven workflow recordings only, reduced to metadata: method, path, status, timing. Balances, PANs, statement bodies and live auth cookies stop reaching S3 — asserted by a test that greps the serialized output for each.

The HAR 1.2 **shape** is preserved and fields are emptied rather than deleted, so a consumer I have not audited cannot break. Request bodies keep their *shape* (top-level field names, plus a ciphertext-length flag) because noui's `detect_unreplayable` needs exactly that to decide browser-vs-replay — an earlier cut dropped bodies entirely and silently disabled that decision, which would have sent ICICI back to the 40-operations-that-all-403 skill.

## Blast radius

**Login recordings are untouched** — full HAR, no outcomes, and rich capture is *inert* rather than merely quiet: `enrich()` bails before invoking its element resolver, so not one line of the evidence path runs. **Workflow recordings that are not browser-driven keep a full HAR**, so replay skills are unaffected.

## Testing

**223 worker + 607 api tests**, 16 suites, green. `pnpm run build` and `pnpm -r run lint` clean.

Includes explicit key-set regression tests on the login bundle, and a dedicated spec for the HAR gating that mocks the capture so a real body reaches `drain()` — asserting against an empty entry list would pass whatever the gate did.

Companion: https://github.com/adoptai/noui/pull/139

---

## Also here now: what a day of live replays against ICICI proved missing

Everything below was found by driving real sessions, not by reading the code. Each
one had already survived review-by-inspection.

**An icon must not disqualify a label** (`ownLabel`). The rule was documented as
"several TEXT-BEARING children" and implemented as "any element child". ICICI's nav
item is `<a><i class="icon"/><span>Credit Cards</span></a>` — two children, one of
which says nothing — so the item was captured with **no text at all**. That cascaded:
no text candidate; and a falsy `ownLabel` also stopped `actionableAncestor` returning
the element, so evidence came from the nav CONTAINER and the click carried the same
positional path as the hover plus a class every submenu item shares. Replay resolved
whichever match came first in the document and failed "not visible" about half of all
runs — while clicking that item BY TEXT works every time. Verified by capture: that
same click now records `text: "Credit Cards"` and a `text` candidate.

**Primitives a compiled skill actually needs**
- `set_checked` can address a control by role and accessible name — a radio styled as
  an image over a hidden input has no unique selector.
- `click_element` can hold a hover through the click. As two commands the menu closes
  in the gap; measured on ICICI, the click was racing the reveal at 1s, 2s, 3s, 4s.
- `get_page_info` reports `ready_state`. A caller deciding whether it may act had only
  the URL, and a form postback re-renders WITHOUT changing it — the portal answered a
  click while still processing the previous one, kept the earlier selection, and
  showed "we are unable to process your request".
- A login `goto` waits for the DOM, not for every last beacon. `load` waits for all 27
  subresources ICICI's login page pulls; it never finished inside the 30s step
  timeout, so the DSL aborted at step 0 and the session never started its health loop
  — leaving a human who signed in over VNC with nothing watching.

**Health and lifecycle**
- A health check that raced a pending navigation is transient, not an auth failure.
- An unevaluable cycle records no verdict at all rather than a wrong one.
- The controller now says WHY a worker is gone. An OOM-killed worker is SIGKILLed
  mid-instruction and logs nothing, so the only symptom was "Worker unreachable: fetch
  failed" — four explanations were chased before someone read the pod status.
- A scaled-up session belongs to the app's tenant, not the caller's: an Admin scaling
  across tenants wrote `undefined` into a NOT NULL column, so every Admin scale-up
  failed.

310 worker tests, 612 API tests.


navjot-adopt:
## Added: the runtime perception fixes (`95c02ee`)

A local ICICI statement-download run failed — 25+ steps, a misclick into "Convert to EMI", then a session kill. Diagnosing it turned up two gaps on the **runtime** side that made the recorder work unable to pay off, so they're folded into this PR.

### `get_page_summary` returned no selectors

Elements came back as bare `{ text, tag }`. No id, no test-id, no aria-label, **nothing to click with but a string**. On a page offering "download", "convert to emi" and "view more transactions", the agent picked wrong — and it could not have done better, because the tool gave it no way to disambiguate.

We had just built a recorder that captures ranked candidates *with match counts*, while the runtime threw all of that away.

Now every element carries the most durable selector that resolves to **exactly one** node, so `click_element` can be used instead of guessing at text. A selector matching several nodes is reported **with its count** rather than presented as an address — that is precisely how the wrong button gets clicked. An element that cannot be addressed carries no selector at all, which beats handing over something that will not resolve.

### Hidden elements were dropped entirely — my own over-correction

I added `.filter(isVisible)` earlier to stop the model chasing HSBCnet's phantom modals. It worked, and it over-corrected: a bank's "past statements" accordion is `display:none` until expanded, so the section the agent needed was invisible and it concluded the portal only offered the current statement. The agent said so itself in the trace:

> *"The page summary tool isn't surfacing the full statement section — likely because it's hidden/collapsed and the tool only shows visible interactive elements."*

Hidden elements now go in a separate `hidden` bucket — in the DOM, not on screen, expand something to reach them. **The main lists keep exactly their old visible-only meaning**, so nothing starts chasing phantoms by default; there's a test pinning that specifically.

### Truncation was silent

The 50-element cap bit without saying so, which reads to the model as "that is the whole page". Now reported.

### Also

Extracted to `page-summary.injected.ts` alongside `dom-recorder.injected.ts` — what the model can see decides what it can do, so it deserves to be readable and testable rather than buried in a switch arm. **It had no behavioural cover at all before this**; there are now 9 tests, written against the two ICICI failures directly.

Worker suite: **232 tests, 17 suites**, green.

### Still open from the same diagnosis

Not in this PR, and worth knowing about:

- **Downloads cannot become operations.** `derive_browser_pages` builds operations from URL transitions, and ICICI's download is an in-page click producing a `blob:` — no URL change, no page, no operation. That is why the build agent hand-wrote `download_credit_card_statement` rather than compiling it. The recorder now captures the evidence (`outcome.download`); the compiler does not yet use it.
- **No parameterisation.** The recording captured one statement period, so "last year" sent the agent hunting for a selector the summary never showed.
- **`navigate` is blocked only by prose.** SKILL.md says never use it; the agent did anyway when stuck, and killed the session.

navjot-adopt:
## Review — deep pass, verified by running the code

Well-reasoned throughout; nearly every fix carries the incident that motivated it. But several headline fixes **do not run on the real path** while their tests pass green.

### Blocking

**1. The revealing hover steals the click's outcome** — `dom-recorder.injected.ts:669` stamps the click, then `:719-720` emits the hover which stamps again at `:793`. The hover gets a *later* `seq`/`event_time` than the click it precedes. `deriveOutcomes` sorts on `event_time` (`recording-outcomes.ts:95-98`) and truncates at the next interaction (`:158-159`), so the click's window is 1ms wide:

```
CLICK  {"navigated":false,"request_count":0,"download":false}
HOVER  {"navigated":true,"to_url":"…/cards","download":true}
```

Hover CARDS → click Credit Cards compiles the click as "did nothing" and credits the hover with the navigation. `dom-recorder.injected.spec.ts:487-490` asserts emission order only, never `seq`. **Fix: stamp the hover first.** This corrupts recordings at the source; everything downstream inherits it.

**2. `browser_driven` never reaches the runner on the warm-pool path** — three bugs in series: `health-server.ts:200` maps a `login` bind to `undefined`; `main.ts:327` then never calls `adoptMode`; `recording-runner.ts:393` returns before the download/popup attach at `:397`. Net: every warm-pool combined recording captures with `richCapture: false` — no candidates, no evidence, no outcomes, no download events. The cold path works; the default path doesn't.

**3. An adopted mode never reaches subframes** — `recording-runner.ts:174` freezes `{rich: this.richCapture}` into `addInitScript`; `adoptMode:407` re-evaluates main-frame only. After a bind-time upgrade every iframe still runs `rich:false`. That is the Finacle-in-an-iframe case exactly.

**4. `set_checked` reports success it did not verify** — `execute-browser-handler.ts:313-318` catches `isChecked()` throwing (which it does for the div-over-hidden-input control this feature was built for) and returns `success:true, checked:true, verified:false`. **NoUI never reads `verified`.** That reinstates "clicked Annual, form stayed on Monthly" one field deeper. Also `:151` accepts `name: ""` without a truthiness check → `getByRole('radio',{name:''}).first()` checks the first unnamed radio and returns `verified: true`.

**5. Health can fail open forever** — `health-predicate-runner.ts:110`: `unevaluable = unevaluableOnly.length > 0 && realVerdicts.length === 0`. PASSing checks are in neither bucket, so with `policy:'any'` overall can be PASS while unevaluable is true, and `keepalive-runner.ts:217-222` returns without writing. A malformed selector is classified unevaluable every cycle (`:413-420`), and nothing watches health staleness — so a signed-out session keeps its last `PASS` indefinitely and keeps serving `/execute/browser`.

### Worth a decision, not just a fix

- **Admin cross-tenant scale is now fail-open** (`sessions.service.ts:125`). Previously died on the NOT NULL constraint; now succeeds, creating a live pod in tenant B running B's login DSL against B's credentials. `role-resolver.helper.ts:13-24` mints Admin per-IdP. The audit row lands under tenant B, so tenant A's log shows nothing.
- **No risky-action classification exists** in these modules. Nothing distinguishes `screenshot` from `click_element` on "Confirm transfer". Two new paths actively defeat human guardrails: `clickThroughOverlays:684` falls back to `dispatchEvent('click')` (bypasses hit-testing), and `set_checked` attempt 2 uses `check({force:true})` (bypasses actionability).

### Medium

- Session-bearing frame URLs leak into responses and stdout (`:126-130`, `:360-372`, `:649-653` → `:82 console.error`); the `bare()` helper at `:122` isn't reused.
- `go_back` bypasses `block_navigate` (`:178-192`) — a full document load, which is what that flag exists to prevent.
- Audit runs before authz/rate-limiting on a global `pg_advisory_lock(42)`, unhandled, and labels `agent_assertion` traffic as `human`.
- `DEFAULT_AUTH_REDIRECT_PATTERN` matches the whole URL including query — `/dashboard?next=/login` → AUTH_FAIL on a healthy session.
- Download attribution is a 3s race (`recording-outcomes.ts:146-149`); the test uses exactly 3000ms. `outcome.download` is a bare boolean with no back-reference, so two files collapse to one. Popups never get `startHarCapture`.
- Forced `match_count = 1` at `dom-recorder.injected.ts:542` fabricates uniqueness the compiler is told to trust.

### Clean

No JS-eval escape — every `evaluate()` passes a fixed function reference, no caller value is ever concatenated. Allowlist is genuinely double-sided. HAR reduction gating is correct and well covered. Redaction happens in-page before the bundle leaves the pod.

### Tests

The pattern: **written against the seam, not the pipeline.** `har-capture.spec.ts:82,88` wires a spy to nothing and asserts it wasn't called — the only check that the peak-memory fix works. `recording-runner.spec.ts:555-566` is titled "never downgrades" and asserts it does.

navjot-adopt:
## Deep review — recorder capture + replay primitives + carried fixes

Reviewed the diff against `origin/dev` across three slices (worker recorder, worker replay handlers, and the non-worker "found along the way" fixes). Grouped by severity; `[confirmed]` = re-read/reproduced. Recorder↔compiler field contract cross-checked against adoptai/noui#139.

### Blocker

- **Migration 034 is not idempotent, and its own guard spec says the column was hand-applied.** `[confirmed]` `034-AddSessionRuntimeError` uses bare `ALTER TABLE "sessions" ADD COLUMN` (035 right beside it uses `ADD COLUMN IF NOT EXISTS`), yet `migrations-registered.spec.ts` states 034 "had already been applied by hand." On that DB, `migrationsRun` executes 034 → `column already exists` → migration txn fails → API pod crash-loops on startup. Make 034 `IF NOT EXISTS` (and `DROP … IF EXISTS`). (`migrations/1708300000034-*.ts:18-24`)

### High

- **Recorder↔types contract is out of sync (functional + typed-consumer risk).** The recorder emits, but `recording.types.ts` does not declare: candidate `kind`s `container_label`/`row_scoped`; `element.frame_url`/`frame_name`; `event_type` `hover`/`dropped`; and event fields `supersedes`/`reveal`/`opened_by_previous`/`reason`. The compiler consumes some of these by name, so any typed consumer / schema validation / column-projection ingest drops exactly the durable-locator and iframe signals these PRs add. Also `role_name` candidate `value` is `"role|name"` while the type doc says it's the accessible name. (`dom-recorder.injected.ts:786,870,894,984` vs `recording.types.ts:31-51,73-90,111`)
- **Login-mode invariant is violated.** Only candidate/element enrichment is gated on `rich`; hover emission, `opened_by_previous`, synthetic script-driven `<select>` change events with `supersedes`, and `dropped` tombstones are emitted in **login** mode too, and `schema_version` is a constant `5` for every bundle. A `login` bundle now changes shape, and the login/HAR compiler (which has regression cover keyed on login shape) can see event types it never saw. Gate them on `rich`, or re-baseline the login contract + bump per-bundle version. (`recording-runner.ts` drain; `dom-recorder.injected.ts:1487-1493`)
- **Leaked listener + interval on teardown/upgrade.** `cleanup()` removes click/input/change/submit and clears `selectTimer`, but never removes the `mouseover` capture listener and never clears the `sweepAbandoned` interval (its id isn't captured). The warm-pool login→workflow path calls cleanup + reinstalls, so each upgrade leaves an extra `mouseover` handler on every pointer move and an extra 5s interval — on the long-lived bank SPAs where it matters. (`dom-recorder.injected.ts:1487,1492,1505-1513`)
- **`go_back` bypasses `block_navigate` and can full-reload cross-origin.** The navigate guard matches only `command === 'navigate'`, but `go_back` calls `page.goBack({timeout})` (default `waitUntil:'load'`); when the previous history entry is a separate document it's a full reload — the refresh that kills the session on a `block_navigate` bank portal. And `go_back` is advertised as the way home *because* navigate is refused. (`execute-browser-handler.ts:49,178-192`)
- **`DEFAULT_AUTH_REDIRECT_PATTERN` is unanchored and now default for every profile.** Matched against the whole URL, so `login.salesforce.com`, `signin.company.com/dashboard`, or `?returnUrl=/login` flip a healthy page to AUTH_FAIL → LOGIN_NEEDED. Tabby is a shared platform; the spec only covers path-position URLs. Anchor to `new URL(url).pathname`. (`health-predicate-runner.ts:26-33,139`)
- **`agent_assertion` exchange borrows an arbitrary user's role.** The exchanged token takes `tenant_id` from the agent payload but `role` from `userRepo.findOne({id: target_user_id})` with no tenant scoping and no agent↔user binding — so an agent token for tenant A can mint `{tenant_id:A, owner_user_id:<any admin>, role:'Admin'}`. Previously capped at `agent_payload.role || 'Operator'`. Scope the role to a user who is a member of the minted tenant. (`auth/token-exchange.service.ts:274-293`)
- **Execute-audit is on the interactive hot path, unguarded, behind a global advisory lock.** `await auditService.log(...)` runs before every `execute/fetch`/`execute/browser`, not try/caught, and `AuditService.log` opens a txn taking `pg_advisory_lock(42)` and re-throws. A transient audit-DB hiccup fails a healthy browser command with a 500; and per-keystroke browser commands serialize on one system-wide lock against all audit writers. Wrap it (logging must not fail the request) and coarsen granularity. (`execute.controller.ts:106-121,158-176`)
- **`getPodRuntime` misattributes the death reason.** It returns the first container with any terminated state whose `reason` is set, without filtering `Completed`/`exitCode 0` or preferring the `worker` container. Under Istio, `istio-init: Completed (exit 0)` or a restarted `novnc` sidecar gets written to `last_runtime_error` and later surfaced by `execute unreachable` as a death cause on a healthy session. (`controller/pod-manager.service.ts:381-401`)

### Medium

- `match_count` for CSS candidates is counted with `document.querySelectorAll`, which misses shadow roots → an element in a shadow root gets `match_count:0` and the compiler drops its best locator (no `0→1` guard like `text` has). (`dom-recorder.injected.ts:730-771`)
- O(n²) text-candidate counting (`allElements` + `indexOf` dedup + full `ownLabel` per node) on every click → capture-time jank on large bank dashboards. Use a Set/Map. (`dom-recorder.injected.ts:827-834`)
- Outcome/download attribution flattens all pages and matches by time only (`page_id` discarded) → a popup navigation/download can be credited to a main-page click. (`recording-outcomes.ts`)
- `page-summary.injected.ts` reports raw `textContent` (account numbers, balances) to the model, while the recorder painstakingly masks PANs — inconsistent. (`page-summary.injected.ts`)
- `set_checked`/`select_option`: label-fallback `label[for=id]` looks up the top page, not the resolved frame; a multi-match selector self-verifies `first()` (reintroducing "clicked-but-wrong-state"); `select_option` never reads state back and double-fires `change`. (`execute-browser-handler.ts:337-343,133-139,391-405`)
- `last_runtime_error` is sticky (never cleared when healthy again) → a one-off sidecar restart's string lingers and mis-explains a much later, unrelated failure. (`controller/reconcile.service.ts:655-665`)
- app-template update authz fails open when the caller's role is `undefined` (`actorRole !== undefined && …`), and the route dropped its `@Roles`. Fail closed. (`app-templates.service.ts:99-108`)
- chunked/compressed large bodies bypass the content-length budget in `har-capture` → the OOM-at-export the change targets still happens for the chunked case. (`har-capture.ts`)

### Low / nits

- `supersedes` rebase uses a single delta; a popup interleaving mid-widget can shift it onto unrelated events. (`recording-runner.ts` ingest)
- 034 `down()` lacks `IF EXISTS`; `req.user.sub` read is dead (jwt.strategy returns `user_id`). (`034-*.ts:24`, `execute.controller.ts:112`)
- worker memory limit 1536Mi→2560Mi lands in the shared `values.yaml` default (fleet-wide, not just local); intended, but call it out. (`charts/browser-hitl/values.yaml:124`)

### Split assessment (reviewability)

Independent of the recorder and worth pulling out for their own review — the auth escalation and the migration idempotency especially:
- **auth** (token-exchange role, namespaced-baton) — its own PR; finding above needs real scrutiny, not a rubber stamp in a 6k-line worker diff.
- **execute-audit** — standalone.
- **controller-observability** (`last_runtime_error` column + writer + reader + 034) — one coherent feature, independent of the recorder.
- **local-dev/ops** (Makefile, values, MTU) — independent; only the shared memory bump has non-local blast radius.

*(Deep review via parallel adversarial passes; the migration blocker was re-verified against the code.)*

apps/api/src/modules/auth/token-exchange.service.ts
```diff
@@ -253,12 +253,37 @@ export class TokenExchangeService {
...
+      throw new UnauthorizedException('target_user_id does not name a known user');
+    }
+
     const ttl = params.requested_ttl_seconds || 3600;
     const jti = crypto.randomUUID();
     const payload: JwtPayload = {
       sub: `federated:${params.target_user_id}`,
       tenant_id: params.agent_payload.tenant_id,
-      role: params.agent_payload.role || 'Operator',
+      role: targetUser.role,
```
rahulbh1510, line 286: BLOCKER: `target_user_id` is a fully client-supplied field on `POST /auth/token-exchange` (see `auth.controller.ts:428-454`) with no tenant restriction, and the `userRepo.findOne` lookup on line 274-275 is deliberately not tenant-scoped (per the test at `token-exchange.service.spec.ts:59-63`). This line now sets the minted token's `role` to the *target user's real role* while `tenant_id` stays the calling agent's own tenant. Concrete exploit: a tenant self-registers an agent client, exchanges for an agent JWT, then calls `/auth/token-exchange` with `target_user_id` = the UUID of an Admin in a *different* tenant (if that UUID ever leaks via logs/support tickets). The result is a token carrying `role: Admin` scoped to the agent's own tenant — and `resolveTenantScope()` returns `undefined` for Admin, meaning every tenant-scoped controller treats it as "no filter, see everything." Before this PR, this branch always set `role: params.agent_payload.role || 'Operator'`, so it could never mint Admin this way. Is `target_user_id` guaranteed to only ever come from a fully-trusted internal broker, never attacker-influenced? If not, this needs a tenant match on the target row or a cap on the role that can be minted here.

apps/worker/src/dom-recorder.injected.ts
```diff
@@ -148,34 +255,990 @@ export function domRecorderScript(): void {
...
+   * amounts, OTP boxes and row indices are what make labels selectable, and
+   * masking those would blind the locators for no privacy gain.
+   *
+   * Masking deliberately happens BEFORE match counting, so two cards whose
+   * labels differ only by number both collapse to the same masked string and
+   * the candidate is scored ambiguous rather than unique. That is the honest
+   * result — the masked locator genuinely cannot tell them apart — and it stops
+   * the compiler from trusting a text fallback that could pick the wrong card.
+   */
+  const maskSensitive = (s: string): string =>
```
rahulbh1510, line 310: BLOCKER: `maskSensitive` only recognizes `[ .-]` as a digit-group separator. Verified by direct execution: `maskSensitive("Card 4315 8105 5762 5005")` (non-breaking-space-grouped — a common bank-UI rendering choice) returns the string **unchanged**; same for comma- or underscore-separated numbers. Several call sites gate on "did masking change it" to decide whether a value is safe to embed raw: `buildRichSelector` (~line 214, aria-label into `selector`), `accessibleName` (~lines 504/524/527, into `element.accessible_name`), and `getDataAttrs` (~line 192, into `data_attrs_json`). Since masking silently no-ops on these separators, the full unredacted card/account number gets embedded into the persisted recording bundle. `normText()` (lines 306-307) is safe because it collapses `\s` (which includes NBSP) before masking — the aria-label/data-attr/selector paths skip that normalization. No test exercises `maskSensitive` against a non-ASCII-space separator.

apps/worker/src/recording-outcomes.ts
```diff
@@ -0,0 +1,195 @@
...
+    // it. The compiled operation then ended at the link, cutting off the period
+    // radio, the year dropdown and the download button: everything that makes
+    // the download a download.
+    //
+    // A label is weak evidence and recency is strong. So take the most recent
+    // click, and let a label win only among clicks that happened alongside it --
+    // which is the case the preference was for: an "Export" finished by a
+    // "Confirm", or a banner dismissed while the file generated.
+    const newest = before[before.length - 1];
+    const tieBreak = before.filter((i) => timed[newest].t - timed[i].t <= LABEL_TIE_MS);
```
rahulbh1510, line 159: CONCERN: `tieBreak` only includes clicks within `LABEL_TIE_MS` (3000ms) of the newest click. Constructed case: `click(0, "Download Statement")`, then an unrelated `click(5000, "Refresh")`, file arrives at `t=12000`. `newest` is the Refresh click; the Download click is 5000ms away so it's excluded from `tieBreak`, `labelled` ends up empty, and attribution falls through to `newest` — Refresh gets `outcome.download: true` while the real download click gets `false`. The existing test in `recording-outcomes.download.spec.ts:38-45` covers this shape only at exactly the 3000ms boundary (inclusive `<=`), so it still passes; a slightly larger, equally realistic gap (downloads can legitimately take up to `DOWNLOAD_WINDOW_MS`=120s per this same file, so an unrelated click 5-10s later while waiting is plausible) is untested and silently misattributed.

packages/shared/src/recording.types.ts
```diff
@@ -13,6 +13,99 @@
...
+ * where it is revisable, and makes ambiguity detectable at compile time: a
+ * candidate with `match_count > 1` cannot be trusted to identify this element,
+ * and one with `match_count === 0` did not even find it.
+ *
+ * `match_count` is -1 when counting failed (an engine that could not evaluate
+ * the selector) — distinct from 0, which is a real "matched nothing".
+ */
+export interface RecordedLocatorCandidate {
+  /** How this candidate addresses the element; also its durability ranking. */
+  kind:
```
rahulbh1510, line 36: CONCERN: This `kind` union does not include `container_label` or `row_scoped`, both of which `dom-recorder.injected.ts` actually emits (`add("container_label", named, -1)` and `add("row_scoped", scoped, -1)`). Since the injected script is untyped (`any`), this ships without a compile error. Any downstream consumer that pattern-matches exhaustively on `kind` will silently drop these two candidates — which, per the in-code ICICI evidence, are emitted frequently. No test in `dom-recorder.injected.spec.ts` covers either kind.

charts/browser-hitl/values.yaml
```diff
@@ -121,7 +121,13 @@ worker:
...
     limits:
       cpu: "1500m"
-      memory: "1536Mi"
+      # Headroom for a long recording. What actually BOUNDS memory is the HAR
+      # body budget (EXECUTE_LIMITS.MAX_HAR_BODY_TOTAL_BYTES); this is margin for
+      # the browser itself plus the transient doubling when the bundle is
+      # stringified at drain. A 1536Mi worker was OOM-killed mid-export on an
+      # ICICI login recording — after the bundle was assembled and before it was
+      # persisted, so the whole capture was lost.
+      memory: "2560Mi"
```
rahulbh1510, line 130: CONCERN (not a defect): this raises the default worker memory limit for staging/production (1536Mi→2560Mi per the PR diff), not just local Kind. It's plausibly justified — the comment above ties it to an OOM incident from HAR bundle stringification during recording drain — but a production capacity change like this is easy to miss buried inside a "workflow recording capture" PR. Was this reviewed against current staging/prod node sizing before merge?

apps/controller/src/pod-manager.service.ts
```diff
@@ -342,6 +342,89 @@ export class PodManagerService {
...
+        ({ ended }) => ended?.reason && ended.reason !== 'Completed' && ended.exitCode !== 0,
+      );
+    // The WORKER's death is the signal this column exists for; a noVNC sidecar
+    // restart is secondary and must not shadow it, and container order is not
+    // guaranteed worker-first. Prefer the container named 'worker'.
+    abnormal.sort(
+      (a, b) => (b.container.name === 'worker' ? 1 : 0) - (a.container.name === 'worker' ? 1 : 0),
+    );
+
+    const picked = abnormal[0];
```
rahulbh1510: Correctness nit: `abnormal` collects any non-Completed/non-zero-exit container across both `containerStatuses` and `initContainerStatuses`, then sorts only to prefer a container literally named `worker` if one is abnormal. If the `novnc` sidecar crashes/restarts while the worker container itself stays healthy, `abnormal` contains only the novnc entry, and `getPodRuntime` reports `terminated: "novnc: <reason>..."` — which `reconcile.service.ts` then writes into `session.last_runtime_error`. A later, unrelated failure in `execute()` (e.g. a transient network blip) then surfaces to the caller as "the worker container terminated: novnc: ...", which is misleading. There's a test for "novnc restart + worker death → worker wins" but none for "novnc restart + worker healthy → should report nothing."

rahulbh1510 (commented):
Two Blockers found: a cross-tenant privilege-escalation path in token-exchange (new — impossible before this PR, since the exchange used to hardcode role from the agent's own payload) and a PAN/account-number masking bypass on non-breaking-space/comma/underscore-separated numbers in the recorder. Four Concerns are inline. Scope note: `Makefile`/`CLAUDE.md` (Kind cluster MTU fixes) and `charts/browser-hitl/values-local.yaml` are unrelated dev-infra changes bundled into this feature PR — low risk, but should have been their own PR. Open question not tied to a line: is `browser_driven` boolean coercion at both `/recording/bind` call sites (`recording-provision.controller.ts:81`, `health-server.ts:209`, both `=== true`) guaranteed to always send an explicit value? The `typeof browser_driven === 'boolean'` check added in `main.ts:332` to distinguish "unspecified" from "false" can never see "unspecified" through this path, since both boundaries collapse it to a concrete `false` first — if any other caller ever omits the field, a bind could silently revert an already-adopted `browser_driven: true` back to `false`. Aikido static scan: 0 open findings for this repo/PR.

apps/worker/src/dom-recorder.injected.ts
navjot-adopt: Fixed in 5f64bba. `maskSensitive`'s separator class is now `[\s.,_-]` (was `[ .\-]`), and `\s` in JS already matches NBSP, thin-space and the other Unicode spaces — so a PAN grouped by any of them, or by comma/underscore, is redacted before it reaches the aria-label / accessible_name / selector / data-attr paths (which mask raw, unlike `normText`). Added tests exercising plain space, NBSP, thin space, comma and underscore.

packages/shared/src/recording.types.ts
navjot-adopt: Added `container_label` and `row_scoped` to the `kind` union in 5f64bba, with a comment noting both carry `match_count: -1` (resolved via Playwright's `:has-text()` engine, not querySelectorAll). No TS consumer in the repo switches exhaustively on `kind` today, but the injected recorder is `any`, so this keeps a future consumer from silently dropping the two kinds the div-nav portals emit most.

apps/controller/src/pod-manager.service.ts
navjot-adopt: Good catch — fixed in 5f64bba. `getPodRuntime` now decides on the worker container first: if it's present and not abnormally terminated, a noVNC sidecar crash/restart is no longer reported (it would otherwise land in `last_runtime_error` and resurface as the phantom cause of a later, unrelated `execute()` failure). A non-worker container becomes the cause only when it explains a *missing* worker (a failed init container, or — if there's no worker container at all — any abnormal one). Added the missing "novnc crashed + worker healthy → reports nothing" test; the existing "novnc restart + worker death → worker wins" still passes.

apps/worker/src/recording-outcomes.ts
navjot-adopt: You're right that the >LABEL_TIE_MS gap was untested, so I pinned it in 5f64bba — but deliberately as a *known limit*, not a behavior change, and here's why. Your Case B (`Download Statement`@0, `Refresh`@5000, file@12000 → recency credits Refresh) is structurally identical to the ICICI case the recency rule was written for and which has a shipped test (`download previous statement` link@0, the real unlabelled `#DOWNLOAD_ESTATEMENT_PDF` button@40000, file@45000): both are labelled-early + unlabelled-late, and they resolve *oppositely*. `looksLikeDownloadControl` reads only `text_content`, and the real ICICI button is unlabelled — so preferring the labelled control in your case would re-credit the link in theirs and reopen the exact bug the recency rule closed. The two are indistinguishable from the only signals attribution has (timestamps + text). The one real disambiguator is a per-click "this action started the download" signal the recorder doesn't yet capture; flagged that in the pinned test's comment as the future fix.

charts/browser-hitl/values.yaml
navjot-adopt: Fair flag on visibility. One thing that de-risks it: only the *limit* moved (1536Mi→2560Mi); the worker `requests.memory` is unchanged at 1Gi. K8s bin-packs on requests, so scheduling density / steady-state capacity is unaffected — what rose is only the per-pod burst ceiling during a HAR drain (the transient doubling when the bundle is stringified, which is what OOM-killed the ICICI capture). The residual risk is many workers bursting concurrently on one node; that's a node-headroom check for whoever owns staging/prod sizing rather than something this PR changes at steady state. Leaving this thread open for that ops confirmation rather than resolving it myself.

apps/api/src/modules/auth/token-exchange.service.ts
navjot-adopt: Confirmed the escalation is real and I'm leaving this thread OPEN for an explicit security decision rather than resolving it. Precise trust boundary: agent clients are `@Roles('Admin')`-provisioned (POST `admin/agent-clients`), not self-serve — so the exploit needs an admin-issued agent secret, not just any attacker. But within that: a *limited* Agent-role credential (scoped to specific `allowed_profiles`) + any leaked Admin UUID (UUIDs aren't secrets) mints `role: Admin` scoped to the agent's own tenant, and Admin → `resolveTenantScope() === undefined` → sees everything in that tenant. So a confined credential can escalate to tenant-wide Admin.

The tension is that the cross-tenant, role-flow-through behaviour is deliberate and test-enshrined (`token-exchange.service.spec.ts` asserts the lookup is not tenant-scoped; an Admin member exchanging through a bot must come back Admin, which is what this line fixed). A hard tenant-match on the target row would close it but break that. Candidate mitigations that preserve the common case: (a) cap the minted role so agent_assertion never mints Admin; (b) require the target user's tenant to equal the minted token's tenant. Both are behaviour/threat-model calls for the security owner — hence not resolving. `target_user_id`'s trust source (fully-trusted internal broker vs. attacker-influenceable) is the deciding question, exactly as you framed it.

rahulbh1510 (approved):

PR #166, opened 2026-08-17
fix(recording): boot warm-pool spares browser-driven so pooled captures match fresh

navjot-adopt:
## Problem
Recording an ICICI browser skill on staging (warm pool) compiled a much poorer skill than the same flow on local (fresh, no pool): pooled recordings captured **~12 interactions vs ~24–28 fresh**, losing the radio/period selection and other clicks. Same worker image (`c390747`) on both — so it wasn't a code version. It's the warm-pool path: a spare **boots lean (`browser_driven=false`, login mode)** and must switch **login→browser at claim** (`adoptMode`), and that switch drops ~half the interactions. Infra's image check confirmed the version is fine; the fault is the boot-lean → switch-at-claim path only pooled pods take.

(The ICICI *download* itself failing is a separate, independent issue — India-only host `mccm.icicibank.com` unreachable from US egress — not addressed here.)

## Fix (2 pieces — keeps the pool's sub-second claim)
1. **`buildPoolAppInput`**: spares boot `browser_policy.browser_driven=true` + `downloads=true`. A browser-driven claim then finds `adoptMode()` a **no-op** (no teardown/reinstall churn) — the recorder + download/popup capture are armed from boot, identical to a fresh pod. Warm-up capture is discarded at bind, so booting rich costs nothing.
2. **`adoptMode()`**: on a **login/api** claim, `browser_driven` flips off (`richCapture` false), and we now **tear the rich listeners back down** (new `detachRichCapture()`, factored out of `detach()`), so a downgraded spare is **byte-identical** to a lean-booted one. Workflow-mode recordings stay rich (`isWorkflow` keeps `richCapture` true), unchanged.

## Non-browser safety
An api/login recording that claims a rich spare downgrades at bind: HAR stays whole (reduction is `browserDriven && isWorkflow`), the bundle carries no workflow-only keys, and the rich listeners are removed — same output as today. Covered by a test.

## Tests
- worker: pooled spare no-op on a browser claim (listeners stay, download captured); clean downgrade for login/api (listeners torn down, bundle login-shaped); lean spare still upgrades. (32 pass)
- api: pool app is created `browser_driven` + `downloads` on. (14 pass)
- Typecheck + full monorepo build green.

## Validation after deploy
Record the ICICI flow **pooled** and check the capture stats: interaction count should now match fresh (~24–28, not ~12).

### DIFF
--- a/apps/api/src/modules/recording/recording-pool.service.ts
+++ b/apps/api/src/modules/recording/recording-pool.service.ts
@@ -290,10 +290,22 @@ export class RecordingPoolService implements OnModuleInit {
         ttl_seconds: 300,
       },
       browser_policy: {
-        downloads: false,
+        // Warm spares boot ALREADY browser-driven so a browser-driven claim needs
+        // NO login->browser mode switch at bind: adoptMode() sees browser_driven
+        // already true and is a no-op, so the DOM recorder + download/popup capture
+        // are armed from boot exactly like a fresh pod. The login->browser switch
+        // is what dropped ~half the interactions on pooled recordings (a claimed
+        // spare captured ~12 clicks vs ~24-28 fresh). `downloads: true` likewise
+        // arms download acceptance from boot. A login/api claim downgrades cleanly
+        // at bind -- adoptMode() flips browser_driven off and tears the rich
+        // listeners back down (see RecordingRunner.adoptMode) -- so its bundle is
+        // byte-identical to a lean-booted spare. Warm-up capture is discarded on
+        // bind either way, so booting rich costs nothing during warm-up.
+        downloads: true,
         clipboard: false,
         file_chooser: false,
         recording_mode: 'login' as RecordingMode,
+        browser_driven: true,
       },
       notification_config: {},
       desired_session_count: this.sizeFor(residential),
--- a/apps/api/src/modules/recording/recording-pool.service.ts
+++ b/apps/api/src/modules/recording/recording-pool.service.ts
@@ -118,21 +118,32 @@ export class RecordingPoolService implements OnModuleInit {
     const appName = this.appNameFor(residential);
     const size = this.sizeFor(residential);
     const tenantId = RECORDING_POOL.SYSTEM_TENANT_ID;
+    const desired = this.buildPoolAppInput(residential);
     const existing = await this.appRepo.findOne({
       where: { tenant_id: tenantId, name: appName },
     });
     if (existing) {
-      if (existing.desired_session_count !== size) {
-        await this.appRepo.update(existing.id, { desired_session_count: size });
+      const needsSize = existing.desired_session_count !== size;
+      // Sync browser_policy on the EXISTING row too. A pool app created before the
+      // browser_driven-spare fix (e.g. a live staging pool) is found here and would
+      // otherwise return early, so its spares keep booting with the old lean policy
+      // and the fix never reaches that environment. Detect the stale row by the one
+      // field that matters and rewrite the policy; only new spares the controller
+      // warms after this pick it up, so drain the pool (or let it rotate) to apply
+      // it immediately.
+      const currentDriven = (existing.browser_policy as { browser_driven?: boolean } | undefined)
+        ?.browser_driven;
+      const needsPolicy = currentDriven !== desired.browser_policy.browser_driven;
+      if (needsSize || needsPolicy) {
+        await this.appRepo.update(existing.id, {
+          ...(needsSize ? { desired_session_count: size } : {}),
+          ...(needsPolicy ? { browser_policy: desired.browser_policy } : {}),
+        });
       }
       return existing.id;
     }
     try {
-      const { app_id } = await this.appsService.create(
-        this.buildPoolAppInput(residential),
-        tenantId,
-        'system:recording-pool',
-      );
+      const { app_id } = await this.appsService.create(desired, tenantId, 'system:recording-pool');
       this.logger.log(
         `Created global ${residential ? 'residential ' : ''}recording pool app ${app_id} (size ${size})`,
       );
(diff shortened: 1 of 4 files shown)


rahulbh1510 (commented):
**Verdict:** Don't merge yet — the fix's mechanism is correct in the worker (`adoptMode`/`detachRichCapture`), but the pool-app config change likely won't reach staging's existing warm pool without a manual step; please resolve that before merging.

**❓ Needs a human decision**
- Is a manual deletion/recreation of the existing global recording pool app row planned as part of this rollout, or should `ensurePoolApp` also sync `browser_policy` on the existing-app branch? As written, the fix is invisible to any environment where that app row predates this deploy.

**🔒 Aikido scan:** No open findings for this PR (scan matched head commit).

apps/api/src/modules/recording/recording-pool.service.ts
```diff
@@ -290,10 +290,22 @@ export class RecordingPoolService implements OnModuleInit {
...
+        // arms download acceptance from boot. A login/api claim downgrades cleanly
+        // at bind -- adoptMode() flips browser_driven off and tears the rich
+        // listeners back down (see RecordingRunner.adoptMode) -- so its bundle is
+        // byte-identical to a lean-booted spare. Warm-up capture is discarded on
+        // bind either way, so booting rich costs nothing during warm-up.
+        downloads: true,
         clipboard: false,
         file_chooser: false,
         recording_mode: 'login' as RecordingMode,
+        browser_driven: true,
```
rahulbh1510, line 321: **🛑 Blocker**

`ensurePoolApp` only sets this `browser_policy` on the create path — if the global pool app row already exists (e.g. in staging, where this bug was originally observed against a live warm pool), `ensurePoolApp` finds it via `appRepo.findOne` and returns early, syncing only `desired_session_count`. `browser_driven`/`downloads` on that pre-existing row are never updated, so spares created under it keep booting with the old lean policy — this fix won't take effect there without a manual deletion/recreation of the pool app row.

apps/api/src/modules/recording/recording-pool.service.ts
```diff
@@ -290,10 +290,22 @@ export class RecordingPoolService implements OnModuleInit {
...
+        // already true and is a no-op, so the DOM recorder + download/popup capture
+        // are armed from boot exactly like a fresh pod. The login->browser switch
+        // is what dropped ~half the interactions on pooled recordings (a claimed
+        // spare captured ~12 clicks vs ~24-28 fresh). `downloads: true` likewise
+        // arms download acceptance from boot. A login/api claim downgrades cleanly
+        // at bind -- adoptMode() flips browser_driven off and tears the rich
+        // listeners back down (see RecordingRunner.adoptMode) -- so its bundle is
+        // byte-identical to a lean-booted spare. Warm-up capture is discarded on
+        // bind either way, so booting rich costs nothing during warm-up.
+        downloads: true,
```
rahulbh1510: **⚠️ Concern**

`browser_policy.downloads` doesn't actually gate `acceptDownloads` for recording sessions — `main.ts`'s `downloadsEnabled = browserPolicy.downloads === true || Boolean(recordingMode)` already forces it true for any VNC recording, pool or fresh, regardless of this flag. Flipping it to `true` here is inert; only `browser_driven` (via `richCapture`) actually controls whether download *events* get captured into the bundle. The comment above implies this flag is doing real work when it isn't.

apps/worker/src/recording-runner.ts
```diff
@@ -596,7 +605,19 @@ export class RecordingRunner {
   }
 
   /** Detach all listeners. Safe to call on SIGTERM and after drain(). */
```
rahulbh1510: **💡 Suggestion**

This doc comment ("Detach all listeners...") now sits above `detachRichCapture()`, which only tears down a subset of listeners — it describes the wrong method. `detach()` below it has no doc comment at all now. Worth moving the comment down to `detach()` and giving `detachRichCapture()` just its own (which it already has, duplicated below).

apps/api/src/modules/recording/recording-pool.service.ts
navjot-adopt: Fixed in 4c89752. `ensurePoolApp` now syncs `browser_policy` on the **existing** row, not just on create: it detects a stale row by `browser_driven` and patches the policy (in the same update as `desired_session_count`). So a pool app that predates the deploy (staging's live pool) gets rewritten. New spares the controller warms after that pick it up; existing warm spares rotate out on claim — drain the pool to apply immediately. Added tests: rewrites a lean row, no-ops an already-driven one (so it's not a write per request).

apps/api/src/modules/recording/recording-pool.service.ts
navjot-adopt: Right — reverted `downloads` to `false` in 4c89752. `main.ts` forces `acceptDownloads` on for any recording via `recording_mode`, so the flag was inert; only `browser_driven` (→ richCapture) gates whether download events reach the bundle. Left a comment saying exactly that so it's not misleading.

apps/worker/src/recording-runner.ts
navjot-adopt: Fixed in 4c89752 — moved the `/** Detach all listeners… */` comment back down onto `detach()`; `detachRichCapture()` keeps only its own (accurate) doc comment.

navjot-adopt:
Re the human-decision question — went with **option 2: `ensurePoolApp` syncs `browser_policy` on the existing-app branch**, so no manual deletion/recreation is needed and the fix reaches any environment whose pool app row predates the deploy (4c89752).

**Rollout note:** the sync updates the pool *app* row, which the controller reads when it warms **new** spares. Spares already warm at deploy time keep the old lean policy until they rotate out (each is single-use — claimed → terminated → refilled). So it self-heals within a pool cycle; to apply immediately, drain the pool once (scale the pool app's `desired_session_count` to 0 and back, or delete the WARM spares).

Also addressed the other two comments (inert `downloads` flag reverted; doc comment moved to `detach()`). All three threads resolved. Worker 32 / API 16 tests green, full monorepo build green.

rahulbh1510 (approved):

PR #173, opened 2026-09-15
fix(worker,api): surface policy-discarded downloads; merge browser_policy instead of replacing it

navjot-adopt:
Two of seven root causes from the ICICI browser-skill RCA (org `87451b06`, 2026-09-15). The member signed in four times and was finally asked to **waive the download** — the only thing they had asked for.

## 1. `list_downloads` could not say the browser had thrown the file away

When `browser_policy.downloads` is false, `main.ts` builds the context **without** `acceptDownloads` and cancels every download as it starts. `list_downloads` then returns `[]` **permanently**, however many times a caller clicks.

An empty list alone is indistinguishable from *"the click did nothing"*. So replay reported

> this step downloaded a file when it was recorded; no file arrived

about a step working exactly as recorded — 12 times, across three replays, before the member waived it.

**Fix:** `listDownloads` returns `disabled_by_policy: true` when the context is not capturing. Playwright exposes no way to read `acceptDownloads` back off a `BrowserContext`, so membership is recorded in a `WeakSet` at the one place it is actually decided — `enableDownloadCapture`.

## 2. `update()` dropped `browser_policy` flags it wasn't asked about

`Object.assign(template, data)` replaced `browser_policy` wholesale. It is a bag of **independent** safety flags written by different callers: the compiler sets `block_navigate`/`downloads` from what a recording proved the app needs; a human sets others from the console. Whichever wrote last silently dropped the rest.

Observed on a browser-driven bank template that **lost `block_navigate`** — after which every `navigate` reloaded the portal and signed the member out mid-task.

**Fix:** `browser_policy` is merged. A caller that genuinely wants a flag off still sends it as `false`; only **absent** keys are preserved.

## Blast radius

Additive for non-harness consumers: `disabled_by_policy` is a new optional field on an existing response, and the merge only changes a PATCH that carries a *partial* `browser_policy` — which previously lost data.

## Tests

- 6 worker specs (`download-capture`), 30 API specs (`app-templates`) — all pass.
- The `browser_policy` merge test **fails without the fix** (verified by stashing it).
- Pre-commit ran `pnpm -r run lint` and a full 7-project build.

## Companions
- **adoptai/noui** `fix/import-keeps-stale-template` — the import that kept the stale template, the goal check that couldn't see a download, and replay's 2.5s download wait.
- **adoptai/adoptai-workflows** `fix/broker-template-reads-and-import-warning` — broker allowlist for template reads, and surfacing the reuse to the agent.


rahulbh1510 (commented):
**Verdict:** Don't block — the diff itself is correct, well-tested, and tightly scoped to its stated intent; nothing here needs to change before merge.

**❓ Needs a human decision**
- `apps/api/src/modules/apps/apps.service.ts:189` (untouched by this PR): `PATCH /apps/:id` still does `...(input.browser_policy !== undefined && { browser_policy: input.browser_policy })` — a full replace, not a merge. This is the exact same failure mode this PR just fixed for templates: an app with `browser_policy = {downloads:true, block_navigate:true, ...}` that gets PATCHed with only `{downloads:false}` will silently lose `block_navigate`. Was this left out of scope deliberately (e.g. a follow-up PR), or should it be fixed alongside this one since it's the same root cause?
- `apps/api/src/modules/app-templates/app-templates.service.ts:228` — `propagateToLinkedApps()` (unchanged by this PR) still does `appRepo.update(app.id, payload)` where `payload.browser_policy` is the template's fully-merged policy, applied as a full overwrite onto every linked app. Since `apps.service.ts` allows `browser_policy` to be set directly on an app independent of its template, an app-level override (e.g. `block_navigate:true` set directly on the app) would be silently wiped the next time anyone edits any field on that app's template. Is template propagation meant to be authoritative over app-level overrides, or should it merge too?

**🔒 Aikido scan:** No open findings for this PR (0 results; scan queried fresh against this run, matched by repo name).

rahulbh1510 (approved):

pratyush0611 (approved):

PR #174, opened 2026-09-16
fix(api,worker): a re-registered template adopts the apps it used to own

navjot-adopt:
## What you saw

Org 87451b06, fresh ICICI build after deleting the older skills and profiles. Auth Manager showed the template as correct:

```json
"browser_policy": {"clipboard":false,"downloads":true,"file_chooser":false,"block_navigate":true}
"created_at": "2026-09-16T09:41:35.950Z"
"updated_at": "2026-09-16T09:41:35.950Z"   // never modified
```

…and its Monitor modal showed **Provisioned apps: 0**. Meanwhile the session it was meant to govern ran with downloads off and cancelled every statement it fetched: `{"downloads": [], "disabled_by_policy": true}`.

Editing the template changed nothing, because the template governed nothing.

## Why

Lookup is by **name**; ownership is by **id**.

| | key |
|---|---|
| find the template for a profile | `profile_name_pattern` |
| link app → template | `template_id` |
| push policy to apps | `template_id` |

`applications.template_id` carries `ON DELETE SET NULL` (migration 020). Deleting the **predecessor** template nulled it on every app it had provisioned. Re-registering under the same name minted a new row with a new id — so the *name* relationship was reproduced and the *id* one permanently lost. `propagateToLinkedApps` filters on `template_id`, matched nothing, and the app kept a month-old policy.

The new template was never deleted. It was born disconnected from the app it was meant to govern.

**Rebuilding a skill is delete-then-register**, so this is the normal path. It is also the mechanism behind the 2026-09-15 run where `block_navigate` was in the template and absent at runtime — which I had wrongly attributed to a console PATCH dropping the field.

## The fix, in two halves

**Forward** — a new `applications.template_pattern`, written whenever an app is provisioned from a template, which the delete cannot erase. On registration a template adopts the apps carrying its pattern and no current template, then runs the existing propagation.

Keyed on the pattern, **deliberately not on a null `template_id`**: after a delete those two are indistinguishable — the entity comment on `template_id` says *"null for manually-created apps"* — so matching a null id would capture apps somebody built by hand and rewrite their `browser_policy`, `login_config` and keepalive underneath them.

**Backward** — the forward half repairs nothing already broken, which on day one is every orphaned app. The link survives on the **profile**: auto-provision sets `profile_id` to the value it looked the template up by, and a delete never touches profiles. So an ACTIVE profile whose `profile_id` matches the pattern names the app that template used to own.

That alone would also match a hand-built app pointed at the same profile id, so the app must *also* carry the name `autoProvisionFromTemplate` gives it (`<template name> <sep> <owner id>`) — only that path produces it. A recovered app is stamped with `template_pattern`, so it is adopted directly next time and never needs the heuristic again.

**So re-registering repairs an existing orphan, not just prevents the next one** — including the icici app in org 87451b06.

## The signal that hid it

A **missing** `browser_policy` and one that says `downloads: false` both fall back to the same defaults in `main.ts` and were reported identically. `list_downloads` now carries `policy_absent` alongside `disabled_by_policy`, and the worker logs the case — so *"this app was never configured"* stops reading as *"this app says no"*, which is what kept this invisible across two full runs.

## Non-harness impact

Additive throughout: a nullable column, a new optional response field, and adoption that can only fire on apps a template provisioned. No existing behaviour narrows, and nothing changes for apps whose template link is intact.

## Testing

`pnpm run test`: 670 api + 326 worker, all 7 projects pass. `pnpm run lint` and `helm lint` clean. Eight new cases cover the re-link, the audit record, the no-op path, a template with no pattern, that adoption never keys on a null `template_id` alone, legacy recovery via the profile, a hand-built app being left alone, and ACTIVE-only profile matching.

The repo's own migration-registration guard caught the unregistered migration before this went up.


rahulbh1510 (commented):
**Verdict:** Don't block — the core fix (adopting orphaned apps back by `template_pattern`, gated on `template_id IS NULL` so hand-built apps are never touched) is sound and well-tested for the forward path; the concerns below are about a schema gap and test-coverage gaps, not correctness defects in the happy path.

**❓ Needs a human decision**
- Is `profile_name_pattern` guaranteed unique per tenant by convention/UI validation, even though the DB has no unique constraint on it? If not, see the concern inline — a coincidental pattern collision between two unrelated templates lets the second one adopt (and overwrite policy on) the first's orphaned apps after the first is deleted.

**🔒 Aikido scan:** No open findings for this PR (scan returned an empty findings list; head commit is `bad615f`).
**✅ CI/Checks:** Could not be checked from this token — confirm on the PR's Checks tab.

apps/api/src/modules/app-templates/app-templates.service.ts
```diff
@@ -50,9 +50,128 @@ export class AppTemplatesService {
...
+   * live template belongs to that template, whatever its pattern says.
+   */
+  private async adoptOrphanedApps(template: AppTemplateEntity): Promise<number> {
+    if (!template.profile_name_pattern) return 0;
+
+    const orphans = await this.appRepo.find({
+      where: {
+        tenant_id: template.tenant_id,
+        template_id: IsNull(),
+        template_pattern: template.profile_name_pattern,
```
rahulbh1510, line 145: **⚠️ Concern**

This join is purely on `template_pattern == template.profile_name_pattern`, and `profile_name_pattern` has no uniqueness constraint in the schema (only `(tenant_id, name)` is unique — see migration `1708300000012-AddAppTemplates.ts`). If a deleted template's orphaned apps carry the same pattern as an unrelated, independently-named template someone creates later, that new template silently adopts them here and overwrites their `browser_policy`/`login_config`/etc via `propagateToLinkedApps`. Worth confirming pattern uniqueness is guaranteed elsewhere (UI validation?), since nothing in the DB enforces it.

apps/api/src/modules/app-templates/app-templates.service.ts
```diff
@@ -50,9 +50,128 @@ export class AppTemplatesService {
...
+      .filter(id => !alreadyFound.includes(id));
+    if (appIds.length === 0) return [];
+
+    const candidates = await this.appRepo.find({
+      where: appIds.map(id => ({ id, tenant_id: template.tenant_id, template_id: IsNull() })),
+      select: ['id', 'name'],
+    });
+    // Mirrors the name autoProvisionFromTemplate builds. Kept as a prefix test
+    // rather than an equality one because the suffix is the owner's user id.
+    const prefix = `${template.name} ${AppTemplatesService.PROVISIONED_NAME_SEPARATOR} `;
```
rahulbh1510: **⚠️ Concern**

The legacy-recovery prefix is built from the *new* template's `name` (`${template.name} — `), but the orphaned app's name was stamped by the *old* (deleted) template's `autoProvisionFromTemplate` call. If a re-registration renames the skill (e.g. "ICICI CC" → "ICICI CC v2") while keeping the same `profile_name_pattern` — a very plausible "fix and rebuild" flow — this prefix check silently fails to match, and the pre-migration orphan is never recovered. The two unit tests for this path use the same string for both `NAME` and `PATTERN`, so this gap isn't covered.

apps/worker/src/download-capture.ts
```diff
@@ -81,6 +81,24 @@ function sanitize(name: string): string {
...
+ * null, so the two cases produce identical behaviour and, until now, identical
+ * reporting. They call for opposite responses: a deliberate `false` is a policy
+ * decision to respect, while a missing policy means the app was never
+ * configured -- which is what an app orphaned from its template looks like, and
+ * what nobody could see while a whole run was spent on it.
+ */
+const policyAbsent = new WeakSet<BrowserContext>();
+
+/** Record that this context's app had no browser_policy to read. */
+export function markBrowserPolicyAbsent(context: BrowserContext): void {
```
rahulbh1510, line 98: **⚠️ Concern**

`markBrowserPolicyAbsent`/the new `policy_absent` field in `listDownloads()` have no test coverage — `download-capture.spec.ts` wasn't touched in this PR, and the existing "flags a context that is not capturing" test never calls `markBrowserPolicyAbsent`, so it can't verify `policy_absent` is set correctly. Same gap on the API side: `credentials.service.ts` now also writes `template_pattern` during auto-provisioning, but `credentials.spec.ts` has no assertion on it.

apps/api/src/modules/app-templates/app-templates.service.ts
```diff
@@ -50,9 +50,128 @@ export class AppTemplatesService {
...
+      },
+      select: ['app_id'],
+    });
+    const appIds = [...new Set(profiles.map(p => p.app_id))]
+      .filter((id): id is string => typeof id === 'string' && id.length > 0)
+      .filter(id => !alreadyFound.includes(id));
+    if (appIds.length === 0) return [];
+
+    const candidates = await this.appRepo.find({
+      where: appIds.map(id => ({ id, tenant_id: template.tenant_id, template_id: IsNull() })),
```
rahulbh1510: **💡 Suggestion**

`where: appIds.map(id => ({ id, tenant_id: ..., template_id: IsNull() }))` builds one OR-branch per id. `where: { id: In(appIds), tenant_id: template.tenant_id, template_id: IsNull() }` expresses the same query more simply and generates a single `IN (...)` clause instead of N OR'd conditions.

apps/api/src/modules/app-templates/app-templates.service.ts
navjot-adopt: Checked, and the answer to your question is **no** — nothing guarantees it. Only `(tenant_id, name)` is unique (migration 012), and `profile_name_pattern` has no constraint, no UI validation I can find, and no check at create.

It is worse than a latent adoption hazard: `autoProvisionFromTemplate` already resolves a template with `findOne({ where: { tenant_id, profile_name_pattern } })`. With a duplicate, **provisioning itself picks arbitrarily** between them — so the assumption is already load-bearing, just unstated.

Two changes:

1. `create()` refuses a colliding pattern, the same shape as the existing name check: *"Template X already uses profile_name_pattern Y — a profile resolves to exactly one template, so it cannot be shared."*
2. Adoption refuses to act at all when a tenant still holds two templates sharing a pattern (rows predating the check). It logs which pattern is ambiguous and what to do.

I deliberately did **not** add a unique index. `migrationsRun: true` means migrations run on API startup, so a unique index over data that already has duplicates would fail the migration and take the API down on deploy — trading a policy bug for an outage. The create-time check gets the guarantee going forward without that risk.

The bias in (2) is deliberate too: a skipped adoption leaves an app on a stale policy, which is recoverable and visible. A wrong one rewrites a working app's `browser_policy` and `login_config`, which is not.

apps/api/src/modules/app-templates/app-templates.service.ts
navjot-adopt: This is the best catch on the PR — you're right, and the two tests using the same string for `NAME` and `PATTERN` is exactly why I didn't see it. A fix-and-rebuild that renames the skill is a normal flow, not a corner, and the orphan would have stayed orphaned with nothing saying why.

The evidence now comes off the **app**, never off the template's current name. `autoProvisionFromTemplate` names it `\`${template.name} — ${ownerUserId}\`` and sets `owner_user_id` to that same owner in the same call, so the name ends with the separator and **the app's own owner id**:

```ts
a.name.endsWith(`${sep} ${a.owner_user_id}`)
```

That suffix does not move when a template is renamed, and it keeps the property the check exists for — a hand-built app still doesn't match unless someone named it in that exact shape *and* set a matching owner.

New test `recovers an orphan even when the rebuild renamed the skill`: template `ICICI CC v2`, app still named `ICICI CC — user-1`, adopted. The two same-string tests stay, since that case is real too.

apps/worker/src/download-capture.ts
navjot-adopt: Both gaps were real. Covered.

`policy_absent` (`download-capture.spec.ts`):
- **separates an app with NO browser policy from one whose policy says no** — asserts `disabled_by_policy` alone before `markBrowserPolicyAbsent`, both flags after
- **never claims a policy is absent on a context that is capturing** — `enableDownloadCapture` wins, since a capturing context had a policy that said yes

`template_pattern` (`credentials.spec.ts`): **stamps the template lineage on the app, id and pattern together**. That matters more than the usual coverage argument — `autoProvisionFromTemplate` is the *only* place `template_pattern` is written, so nothing else can make the adoption path true. Untested, the whole forward fix could quietly be a no-op.

674 api + 328 worker passing.

apps/api/src/modules/app-templates/app-templates.service.ts
navjot-adopt: Taken — `In(appIds)` now, one `IN (...)` instead of N OR'd branches. The array form was left over from an earlier shape where each branch carried different conditions; it doesn't any more.

rahulbh1510 (approved):

PR #176, opened 2026-09-16
fix(worker): stop decoding .xlsx/.docx as text on /execute/fetch

siddhant-adoptai:
## What

`/execute/fetch` returned every .xlsx, .docx and .pptx as UTF-8 text instead of base64, irreversibly corrupting the file.

## Why it happened

Both text-vs-binary branches classified responses with `contentType.includes('xml')`. OOXML media types are named `application/vnd.openxmlformats-officedocument...`, so they matched the substring — but the payload is a ZIP archive, and a UTF-8 decode maps every invalid byte sequence to U+FFFD, destroying the bytes rather than re-encoding them.

## Why it went unnoticed

The failure is silent. Measured against a real Workday report export served as `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet;charset=UTF-8`:

- 7,633 U+FFFD replacement characters
- `zipfile` → *Bad offset for central directory*
- response still carried HTTP 200, the correct `content-disposition` filename, `encoding: "utf-8"` and `truncated: false`

Nothing downstream has a signal that it is holding a corrupt file.

It also suppressed a working feature: the platform's WDL executor persists binary bodies to object storage and returns a presigned download URL, gated on `encoding === "base64"`. Because the worker reported `utf-8`, that path never ran for a spreadsheet.

## The fix

Per RFC 7303 and RFC 6839, an XML media type is `text/xml` or `application/xml` — with their `-dtd` and `-external-parsed-entity` variants — or any subtype carrying the `+xml` structured-syntax suffix. Matching that instead of the bare substring keeps `image/svg+xml`, `application/soap+xml` and friends textual, while OOXML falls through to the base64 branch it always belonged in.

The predicate is extracted into one exported `isTextualContentType`, so the two call sites cannot drift apart and the logic is directly testable. It now also parses the media type out of the header before matching, rather than testing the raw value with its parameters attached.

## Compatibility

Swept 672 header variants (24 media types x 7 parameter forms x leading whitespace x casing) through the old and new predicates.

**binary -> textual: only `text/html` and `text/plain`, and only with leading whitespace** (the old `startsWith('text/')` failed where the new one trims). Both are genuinely text, so no new corruption path is introduced anywhere.

**textual -> binary: 12 media types, in three groups**

1. The bug being fixed — xlsx, docx, pptx, OOXML templates.
2. Correctly reclassified — `application/vnd.wap.wbxml` (WAP *Binary* XML), and `application/pdf`, `application/zip`, `image/png`, `video/mp4`, `font/woff2`, `application/vnd.ms-excel` **when a parameter contains json/xml/svg**, e.g. `application/pdf; filename="b.xml"`. The old code matched against the raw header, so it was testing the filename parameter — a PDF with an `.xml` filename was being destroyed. Same bug class, second instance, also fixed.
3. Non-standard spellings `application/x-xml` and `application/xml+rss`, neither IANA-registered, which will now be base64'd. Recoverable (base64 preserves bytes), unlike the failure being fixed.

## Downstream

Nothing inside tabby consumes `encoding` — it is produced only. python-mcp and adoptwebui have no consumers. ProjectA3's `widdle_executor` is the sole consumer; its `test_tabby_routed_rest.py` suite passes 32/32, including `test_persist_binary_download_uploads_and_returns_reference` and `test_route_rejects_binary_download_for_multicall_ops`.

**Worth briefing:** a `REST_LOOP` / `PAGINATION` / `REST_DELAY` step fetching an Office file previously "succeeded" with a mangled string, and will now hit the multi-call guard and fail explicitly. Silent-wrong becomes loud-correct, but it is a new failure surface.

## Testing

- 60 lines of new unit tests: OOXML bare and with charset, pdf/zip/octet-stream/png, plus the XML types that must stay textual (`application/xml-dtd`, `application/soap+xml`, `image/svg+xml`, uppercase variants)
- Verified the tests are not vacuous — the old predicate returns `textual: true` for all three OOXML types
- `nx test worker`: 352 passed, 21 suites. Full monorepo suite green. Lint and build clean.
- End-to-end on a real 2,000-row xlsx: old path inflated 47,614 bytes to 86,302 and produced `BadZipFile`; new path round-trips byte-identical (matching sha256) and opens with 2,001 rows.


apps/worker/src/execute-handler.ts
```diff
@@ -176,19 +198,11 @@ export function registerExecuteHandler(app: Express, page: Page): void {
...
-            contentType.includes('json') ||
-            contentType.includes('xml') ||
-            contentType.includes('javascript') ||
-            contentType.includes('x-www-form-urlencoded') ||
-            contentType.includes('svg');
+          // bodies go through resp.text() as before; anything else is read as
+          // raw bytes and base64-encoded so binary payloads (e.g.
+          // application/pdf, .xlsx) survive transit intact instead of being
+          // mangled by a UTF-8 text decode.
+          const isTextual = isTextualContentType(respHeaders['content-type'] || '');
```
rahulbh1510: **🛑 Blocker**

`isTextualContentType` is a Node-module import, but this line runs inside the `page.evaluate()` callback (the same-origin path), which Playwright serializes and executes in the browser's JS realm — that identifier won't exist there. This repo already documents the rule this violates (`page-summary.injected.ts`: "must stay self-contained: no outer closures, no imports"). Every same-origin fetch will throw `ReferenceError` here, get swallowed by the outer `.catch`, and silently fall back to `fetchViaContext()` — which skips the page's JS context, so any per-request token-minting interceptor this path exists to preserve no longer runs, and every same-origin call now pays for a doomed round-trip first. None of the 352 passing tests catch this since the new spec only calls `isTextualContentType` directly in Node.

rahulbh1510 (commented):
**Verdict:** Blocking — the classification fix itself is correct and well-tested, but the refactor breaks the same-origin `/execute/fetch` path at runtime; see the inline comment.

**🔒 Aikido scan:** No open findings for this PR.
**✅ CI/Checks:** Could not be checked from this token (Checks API not available to it) — confirm on the PR's Checks tab.

siddhant-adoptai:
Good catch — confirmed, and fixed in 83016bf.

The blocker is real and my refactor caused it. `isTextualContentType()` was being called from inside the `page.evaluate()` callback, which Playwright serializes into the browser realm where the Node import does not exist. Verified in a real Chromium, not by reading:

```
BEFORE -> page.evaluate: ReferenceError: isTextualContentType is not defined
AFTER  -> ran in page: {"xlsx":false,"svg":true,"html":true,"pdf":false}
```

Your read of the consequence is exactly right: the `ReferenceError` was swallowed by the `.catch` below it, so the symptom was not a failure but a silent downgrade to `fetchViaContext()` — losing the per-request token-minting interceptors that path exists for, after a wasted round-trip. The repo already documents this constraint in `page-summary.injected.ts` and `dom-recorder.injected.ts`; I violated it.

**Fix:** the callback now carries its own self-contained copy of the predicate. I took the inline option rather than classifying in Node because the in-page path uses `resp.text()`, which honours the response charset — returning raw bytes and decoding UTF-8 in Node would silently mis-decode non-UTF-8 same-origin responses. That would trade this bug for a quieter one.

**On "no test catches this"** — also correct, and that was the more important half of the review. Two tests now hold it:

1. `never calls the module-scope isTextualContentType from the browser realm` — slices the `page.evaluate` callback out of the source, strips comments, and asserts its executable code never references the module-scope helper. This fails on the exact diff you blocked.
2. `in-page predicate stays in lockstep with the exported one` — extracts the inlined copy from source, reconstructs it in Node, and asserts it agrees with the exported predicate across 19 content types (OOXML with and without charset, the `filename="a.xml"` param case, wbxml, the real XML types, json/html/form, empty).

The duplication is forced by the realm boundary rather than chosen, so those two tests are what keep the halves honest.

Worker suite: 374 passed, 21 suites (was 352; +22 from the new tests). Lint and full build clean.

Re CI: agreed, worth confirming on the Checks tab before merge.

apps/worker/src/execute-handler.ts
```diff
@@ -176,19 +198,30 @@ export function registerExecuteHandler(app: Express, page: Page): void {
...
+          // mangled by a UTF-8 text decode.
+          //
+          // Deliberately duplicated from isTextualContentType(). This callback is
+          // serialized by page.evaluate() and runs in the browser realm, where a
+          // Node import does not exist — calling the shared helper here throws
+          // ReferenceError, and the catch below would silently fall back to
+          // fetchViaContext(), losing the in-page interceptors this path exists
+          // for. Same constraint as page-summary.injected.ts. The spec fails if
+          // this copy and the exported one ever disagree.
+          const isTextualInPage = (ct: string): boolean => {
```
rahulbh1510, line 213: **💡 Suggestion**

Nice fix — the callback is now fully self-contained (string/regex primitives only) and behaviorally verified against the exported predicate via the `execute-handler.spec.ts` extraction test. One minor fragility: that test locates the callback by matching the exact literal `      ).catch(async (err: Error) => {`. If this ever gets reformatted (e.g. a future Prettier pass changes the wrapping/indentation), `indexOf` silently returns `-1` and the slice degrades instead of failing loudly — worth a comment noting that assumption, or an explicit assertion that both markers were found (the "locates the page.evaluate callback" test only checks length/content, not that both indices resolved).

rahulbh1510 (commented):
**Verdict:** Unblocked — the previous blocker is fixed correctly. Approving the reasoning here as a comment-only review per my scope; no new blockers found.

**What I checked:** the in-page callback (`apps/worker/src/execute-handler.ts:213`) now defines `isTextualInPage` as a fully self-contained copy — same media-type parsing, same inlined XML regex (not a reference to the module-scope `XML_MEDIA_TYPE` constant either) — so nothing it touches crosses the Node/browser realm boundary anymore. The unchanged cross-origin path (`fetchViaContext`, line 139) still correctly calls the shared exported `isTextualContentType`, so no unnecessary duplication was introduced where the realm boundary doesn't apply.

Verified the two new regression tests actually enforce what they claim: the "never calls the module-scope helper" test greps the callback's de-commented source for `isTextualContentType(` (would fail immediately if this were reverted to a shared call), and the "stays in lockstep" test extracts the in-page copy via regex, evals it with `new Function`, and diffs its output against the real export across 19 content types — including the WAP-binary-XML and PDF-with-xml-filename-param edge cases from the original PR. Both are sound, if a little unusual (regex-based source extraction rather than an AST) — flagged one minor fragility inline as a suggestion, not a blocker.

**🔒 Aikido scan:** No open findings for this PR (re-checked against the new head SHA).
**✅ CI/Checks:** Could not be checked from this token — confirm on the PR's Checks tab.

apps/worker/src/execute-handler.ts
siddhant-adoptai: Fixed in 83016bf. The in-page call site now defines a self-contained `isTextualInPage` (inlined regex, no reference to the module-scope `XML_MEDIA_TYPE` either), so nothing in the callback crosses the realm boundary; the cross-origin path at L139 still uses the shared export. Two tests added: one greps the de-commented callback source for `isTextualContentType(`, one extracts the in-page copy and diffs it against the export across 19 content types. Verified in a real Chromium before/after — ReferenceError before, correct classification after.

apps/worker/src/execute-handler.ts
siddhant-adoptai: Addressed in 289e695. The end marker no longer carries its indentation, and both indices are now asserted at module load — a missing marker throws with both offsets rather than slicing the wrong span. 374 tests pass.

navjot-adopt (approved):

rahulbh1510 (approved):

PR #178, opened 2026-09-17
feat(worker): stream large downloads to an object store instead of inlining them

siddhant-adoptai:
## What

A file larger than `MAX_RESPONSE_BODY_BYTES` (5MB) cannot leave the worker at all. Both `/execute/fetch` and `/execute/browser` `get_download` return bytes as base64 inside the JSON response, so the file is base64'd (+33%), parsed and re-serialised by the API on the way through. This adds an object-store sink to both paths.

## Why the cap is not raisable

- A 200MB file is **267MB of base64**, held simultaneously by worker, API (limit **1Gi**) and caller.
- Past ~384MB raw it exceeds V8's maximum string length outright (`0x1fffffe8`).
- The worker's own 2560Mi limit is shared with Chromium, and the comment on that limit records a real OOM-kill at 1536Mi caused by exactly this transient doubling.

## The two sinks

| path | ceiling | memory |
| --- | --- | --- |
| `/execute/fetch` with `upload_url` | `MAX_SINK_BODY_BYTES` (256MB) | buffered once — Playwright's `APIResponse` has no streaming accessor |
| `/execute/browser` `put_download` | none | flat — `createReadStream` → PUT, never materialised |

Both take a **caller-minted presigned PUT**, so the worker holds no object-store credentials: the URL is scoped to one key, one method, one expiry. Both return `{size_bytes, sha256, upload_status}` in place of the bytes. The PUT itself lives in one shared `presigned-upload.ts` so the two cannot drift in how they validate a URL, set Content-Length, or report a rejection — the same reasoning that made the content-type predicate a single exported helper in #176.

`download_url` is added alongside, because `navigate` cannot trigger a download: Chromium hands a `content-disposition: attachment` response to the download manager and **aborts the navigation**, so `page.goto` rejects with `ERR_ABORTED` for a download that in fact succeeded.

## Failure modes closed

Each of these would otherwise produce a stored file that looks fine and is not:

- **An auth wall is never stored.** An expired portal session answers a document URL with 200 and an HTML login page. Nothing is uploaded unless the response is marked `content-disposition: attachment`; the page comes back inline instead so the caller can see why. `upload_always: true` covers a server that sends no disposition header.
- **A non-2xx upstream is never stored.**
- **Over the sink limit fails with 413**, naming `put_download` as the route with no ceiling — rather than truncating into a short object nobody discovers is broken until they open it.
- **The API dropped the sink fields entirely.** The worker payload is assembled field by field, so `upload_url` never reached the worker: a caller could request a sink, get a capped inline response, and have nothing anywhere report the downgrade. The sink could not have worked in a deployed stack before this.
- **Timeouts.** `MAX_TIMEOUT_MS` sizes an API call; a 200MB export needs ~27Mbps sustained to fit in 60s. `MAX_SINK_TIMEOUT_MS` (5 min) applies at both hops, and only to a sink request or `put_download`.
- **Disk.** `MAX_DOWNLOADS` bounds the count, which sufficed while nothing over 5MB could be captured usefully. Twenty retained exports at these sizes is several GB of ephemeral disk, and an overrun is answered by the kubelet evicting the pod, not by an error anyone sees. Retention is now bounded by bytes as well.
- On upload failure the local copy is **kept**, so a caller whose presigned URL expired can mint a fresh one and retry the same id; on success it is released.

## Testing

- Build green, lint clean, **681 API / 342 worker / 103 shared** tests.
- Every new guard verified non-vacuous — disabled individually, confirmed the specific test fails, restored.
- The eviction test uses sparse files; an earlier version allocated 1.4GB of real buffers per run.

## Relationship to #176

Branched from the **same `dev` commit (`07c0729`)** as #176, and does not contain it. Both touch `execute-handler.ts`. Merged locally against #176 in a scratch worktree: **auto-merges clean, builds, 388 worker tests pass, all 7 projects green**. Either can merge first.

Worth noting the two are complementary rather than overlapping: #176 makes the *inline* path return correct bytes; this gives anything too large for that path a way out. The sink's attachment check is independent of the content-type predicate, so it behaves the same either side of #176.


rahulbh1510 (commented):
**Verdict:** Block — two constructible failure paths in the new object-store sink undercut the PR's own stated safety goals: the `/execute/fetch` sink buffers past its own 256MB cap before checking it (OOM risk, not just theoretical — this is the exact failure mode the PR is meant to close), and `uploadToPresignedUrl` uses an unproxied `fetch()` that bypasses this codebase's documented egress-allowlist convention with no host-level compensating control.

**❓ Needs a human decision**
- What's the actual trust boundary on `upload_url`/`upload_headers`? If it's always minted by fully-trusted internal orchestration (never influenced by end-customer WDL config or an LLM-driven agent), the egress-proxy bypass is lower-stakes than I'm treating it — but nothing in the code or PR description states or enforces that assumption.
- Is a private-IP/link-local host block (in addition to the scheme check) an acceptable stopgap for `put_download`'s streaming case, given Playwright's request API can't stream and can't go through the proxy as-is?

**🔒 Aikido scan:** No open findings for this PR (0 results from the code-quality findings endpoint). Note: the response didn't expose a scanned-commit field, so I couldn't independently confirm the scan ran against this exact head SHA.
**✅ CI/Checks:** Not checked — this token cannot call GitHub's Checks API (documented limitation, not attempted).

**Also verified (not part of the checklist findings above):** the PR's own #176-compatibility claim. Tabby's `dev` has since actually merged #176 (commit `2262afb`), so I merged this PR's head against the *current* `dev` tip (not just #176's branch, which is what the author's scratch-worktree check used) — `git merge --no-commit` auto-merges cleanly, only `execute-handler.ts` needed auto-merging, no conflicts. I could not run the build or test suite (no Node/pnpm available in this environment) to confirm "388 worker tests green" independently.

apps/worker/src/execute-handler.ts
```diff
@@ -138,6 +154,107 @@ export function registerExecuteHandler(app: Express, page: Page): void {
...
+        // the report and does not announce itself as anything else.
+        if (!isAttachment && !body.upload_always) {
+          return skip(
+            `response is not an attachment (content-type ${contentType || 'unknown'}, no ` +
+              'content-disposition) — it is more likely an auth wall or an error page than ' +
+              'the file. Pass upload_always:true to store it anyway.',
+          );
+        }
+
+        const buf = await resp.body();
```
rahulbh1510, line 241: **🛑 Blocker**

`MAX_SINK_BODY_BYTES` is checked only after `resp.body()` has already fully materialized the response in memory — so an upstream larger than 256MB (a bigger-than-expected export, or a caller-controlled `url` that just serves a huge body) is buffered in full into the worker's 2560Mi pod, shared with Chromium, before the 413 ever fires. That's the exact OOM shape this PR's own reasoning describes (`MAX_SINK_BODY_BYTES` comment: "bounds that against the pod's 2560Mi"). `get_download`'s existing size check (via `fs.stat`, before reading) checks size before materializing; this path doesn't.

apps/worker/src/presigned-upload.ts
```diff
@@ -0,0 +1,100 @@
...
+    hash.update(source);
+    body = new Uint8Array(source);
+  } else {
+    source.on('data', (chunk) => hash.update(chunk));
+    body = Readable.toWeb(source) as ReadableStream;
+  }
+
+  let resp: Response;
+  try {
+    resp = await fetch(uploadUrl, {
```
rahulbh1510, line 155: **🛑 Blocker**

This PUT goes through Node's global `fetch()`, not `context.request` (Playwright's `APIRequestContext`) — the established pattern this same codebase uses everywhere else specifically because it "inherits the browser's proxy" (see `health-predicate-runner.ts:168`, and the comment a few lines above in `execute-handler.ts`). That proxy is the per-session egress-allowlist; there's no Kubernetes NetworkPolicy for the `worker` component (`network-policies.yaml` only covers api/controller/postgres/redis/nats), so it's the only thing constraining worker egress today. `validateUploadUrl` checks scheme only, not host — a caller-supplied `upload_url` can point at any internal service or a cloud metadata endpoint, and on a non-2xx the target's response body (up to 200 chars, line 95) is echoed back in the thrown error, which is a read oracle into whatever it reached. For the `/execute/fetch` sink this is fixable (the body is already a fully-buffered `Buffer` at the call site, so it could route via `context.request.fetch`); `put_download`'s streaming case is harder since Playwright's request API doesn't take a stream — but that tension isn't mentioned anywhere in the PR.

apps/worker/src/execute-handler.ts
```diff
@@ -138,6 +154,107 @@ export function registerExecuteHandler(app: Express, page: Page): void {
...
+
+        /** Return the body inline (bounded) and say why nothing was stored. */
+        const skip = async (reason: string): Promise<ExecuteFetchResponse> => {
+          const buf = await resp.body();
+          const wasTruncated = buf.length > maxResponseBytes;
+          const shown = wasTruncated ? buf.subarray(0, maxResponseBytes) : buf;
+          return {
+            status,
+            headers: respHeaders,
+            body: shown.toString('utf-8'),
```
rahulbh1510: **⚠️ Concern**

`skip()`'s returned body does `buf.toString('utf-8')` unconditionally, regardless of content type. This is the same bug class PR #176 fixed for `fetchViaContext` in this exact file via `isTextualContentType()` (blind UTF-8 decode maps invalid byte sequences to U+FFFD, silently destroying binary content). If an upstream returns a 200 binary response without `content-disposition: attachment` and without `upload_always:true`, the diagnostic body shown to the caller here will be mangled. Doesn't corrupt the actually-uploaded file (that path uses raw `buf`), only the "here's why nothing was stored" fallback text.

apps/worker/src/presigned-upload.ts
```diff
@@ -0,0 +1,100 @@
...
+      method: 'PUT',
+      headers: {
+        'Content-Type': contentType || 'application/octet-stream',
+        'Content-Length': String(sizeBytes),
+        ...(extraHeaders || {}),
```
rahulbh1510: **⚠️ Concern**

`extraHeaders` (caller-supplied `upload_headers`) is spread last, so it can override the `Content-Type`/`Content-Length` computed just above from the real byte count — e.g. `upload_headers: {"Content-Length": "1"}` sends a declared length that doesn't match the actual body, which object stores generally either reject or (worse) use to silently truncate what gets stored. Worth stripping `content-type`/`content-length` (case-insensitively) from `extraHeaders` before the spread, or spreading it first so the computed values always win.

apps/worker/src/presigned-upload.ts
```diff
@@ -0,0 +1,100 @@
...
+  cmd: string,
+  extraHeaders?: Record<string, string>,
+): Promise<PresignedUploadResult> {
+  const hash = createHash('sha256');
+  let body: BodyInit;
+  if (Buffer.isBuffer(source)) {
+    hash.update(source);
+    body = new Uint8Array(source);
+  } else {
+    source.on('data', (chunk) => hash.update(chunk));
```
rahulbh1510: **💡 Suggestion**

Attaching `.on('data', ...)` before `Readable.toWeb(source)` forces the stream into flowing mode ahead of the web-stream adapter taking over — a known footgun area in Node stream composition. `download-capture.spec.ts`'s `put_download` tests exercise this with real fs streams and assert byte-for-byte equality, which is reassuring, but I couldn't run the suite myself in this environment (no Node available) to independently confirm there's no edge case (e.g. very small or zero-byte files) where a chunk is dropped or duplicated. Worth a deliberate second look given how load-bearing the sha256/byte-count correctness is here.

apps/worker/src/execute-handler.ts
siddhant-adoptai: Correct, and this was the important one — thank you. The limit existed to stop the response being held in memory, and checking it after `resp.body()` meant materialising the whole thing to decide it was too big to hold. Fixed in `11be0da`: the declared `content-length` is now refused before a byte is read.

One thing I want to be straight about rather than imply it is fully closed: a chunked response declares no length, and Workday's document endpoint — the actual use case here — is chunked. Playwright's `APIResponse` exposes no streaming accessor, so there is nothing to check against mid-read, and that case still buffers. The post-read check stays as a correctness backstop, not a memory one, and the comment now says so. `put_download` is the path with no such exposure, since it streams to disk.

Test: `refuses an oversized response on content-length, before reading the body` asserts the body function is never called.

apps/worker/src/presigned-upload.ts
siddhant-adoptai: Agreed on the substance, with one constraint worth stating: `context.request` cannot stream a body, so routing the PUT through it would force `put_download` to buffer a whole export — the exact thing it exists to avoid. So the mechanism stays, and the guard moves to where it can work.

Fixed in `11be0da`: `validateUploadUrl` now resolves the host and rejects private, loopback, link-local (including `169.254.169.254`), CGNAT, unique-local and reserved addresses — every resolved address, not just a literal, so a name that resolves inward is refused too. It mirrors the platform's own `_validate_tabby_target_url`.

The read oracle is closed separately: the store's response body is no longer returned to the caller at all. The status is what a caller acts on; the body is logged for whoever is debugging a mis-signed URL.

Tests: `refuses an internal upload_url before fetching anything` (metadata IP, loopback, RFC1918, IPv6 `::1`) and `does not echo the object store response body back to the caller`.

apps/worker/src/execute-handler.ts
siddhant-adoptai: Fixed in `11be0da`. You are right that it is the same bug class the sibling path just had — a UTF-8 decode maps every invalid byte to U+FFFD, destroying the bytes rather than re-encoding them.

`skip()` now picks the encoding from the content type: an HTML auth wall (the common reason to be here) still comes back as readable text, and a skipped binary — a non-2xx carrying a PDF, say — comes back as base64 with `encoding: "base64"` rather than mojibake.

Two tests pin both directions: `base64s a skipped binary instead of mangling it with a UTF-8 decode` (asserts byte equality through the round trip) and `still returns a skipped HTML auth wall as readable text`.

apps/worker/src/presigned-upload.ts
siddhant-adoptai: Fixed in `11be0da` — caller headers are now spread first so the computed ones win.

Worth naming why this mattered more than it looks: a caller-supplied `Content-Length` that disagrees with the body is exactly how a short object gets stored under a full-looking size, which is the failure mode this whole PR exists to prevent. `Content-Type` matters too, since it is what the store hands back to whoever downloads the file.

Unrelated headers (`x-amz-*` and friends) still pass through untouched — the test asserts all three behaviours at once.

apps/worker/src/presigned-upload.ts
siddhant-adoptai: Good catch to raise, and I went and checked it properly rather than reasoning about it.

Changed in `f58ca25`: the hash now runs inside the pipeline as a `Transform` rather than from a `data` listener, so every chunk is hashed exactly once, in order, on its way to the upload — the same bytes by construction rather than by timing.

**But I want to be accurate about what that is: no defect was reproduced.** I added the size-boundary cases you were unable to run (zero-byte, single-byte, sub-chunk, exactly one 64KiB chunk, multi-chunk at 3×64KiB+17, each filled with a pseudo-random pattern so a dropped or duplicated chunk cannot coincidentally hash equal) — and **all five pass against the previous implementation too.** `Readable.toWeb()` takes the stream over synchronously, before flowing actually begins, so the footgun does not fire here.

So this removes the reliance on that timing rather than fixing an observed bug. The tests stay regardless, since sha256 and byte-count correctness is load-bearing for everything built on top of this.

apps/worker/src/presigned-upload.ts
```diff
@@ -0,0 +1,187 @@
...
+        cb(null, chunk);
+      },
+    });
+    source.on('error', (err) => hashing.destroy(err));
+    body = Readable.toWeb(source.pipe(hashing)) as ReadableStream;
+  }
+
+  let resp: Response;
+  try {
+    resp = await fetch(uploadUrl, {
```
rahulbh1510, line 155: **⚠️ Concern**

`fetch(uploadUrl, ...)` uses the default `redirect: 'follow'`. Since `upload_url` is attacker-controlled, a caller can point it at a public server they own that responds 307/308 with `Location: http://169.254.169.254/...` (or any internal address) — undici follows automatically and re-issues the PUT there, fully bypassing the `isBlockedAddress` check above (which only validated the original host's resolved addresses, not the address actually connected to post-redirect). Impact is blind (the success path never reads `resp.text()`, so nothing is echoed back), but it's still request forgery against internal endpoints. Consider `redirect: 'manual'` and re-validating the `Location` before following.

apps/worker/src/presigned-upload.ts
```diff
@@ -0,0 +1,187 @@
...
+export interface PresignedUploadResult {
+  size_bytes: number;
+  sha256: string;
+  upload_status: number;
+}
+
+/**
+ * Is this address one the worker must never be aimed at?
+ *
+ * Mirrors the platform's own via:tabby target guard. Cloud metadata lives on a
```
rahulbh1510, line 33: **❓ Question**

Couldn't find `_validate_tabby_target_url` (or any equivalent resolved-address guard) anywhere in tabby, adoptwebui, python-mcp, or adoptai-workflows — grepped all four. Is this function in a repo I don't have visibility into, or was "mirrors the platform's own guard" describing intent rather than an existing, verified pattern? The logic itself (private/loopback/link-local/CGNAT/reserved ranges, IPv4 + IPv6) looks sound on its own merits either way.

apps/worker/src/presigned-upload.ts
```diff
@@ -0,0 +1,187 @@
...
+    throw new Error(`${cmd}: upload_url scheme "${parsed.protocol}" not allowed. Use http: or https:`);
+  }
+  const host = parsed.hostname.replace(/^\[|\]$/g, '');
+  if (!host) {
+    throw new Error(`${cmd}: upload_url has no host`);
+  }
+
+  const candidates = new Set<string>([host]);
+  try {
+    for (const { address } of await dnsLookup(host, { all: true })) {
```
rahulbh1510, line 96: **💡 Suggestion**

`validateUploadUrl` resolves and checks DNS once, but the later `fetch()` call re-resolves independently when it actually connects — a low-TTL DNS record could answer safe here and blocked at connect time (rebinding). Lower priority than the redirect issue since it needs attacker control of DNS timing, but worth a comment noting it's accepted, if it is. Also: the new DNS-resolution branch itself has no test — the added tests only cover literal blocked IPs (169.254.169.254, 127.0.0.1, etc.), not a hostname that resolves to one.

rahulbh1510 (commented):
**Verdict:** Follow-up verification of the two blockers from the previous review (commits `11be0da`, `f58ca25`) — both are substantially fixed, but the SSRF fix has a redirect-based bypass that should be closed before this is fully closed out. Doesn't have to block merge on its own if the team accepts blind-SSRF-via-redirect as a tracked follow-up, but flagging it as a real gap rather than closed.

**Verified resolved:**
- OOM ordering (`execute-handler.ts`): declared `content-length` is now refused before `resp.body()` is called — confirmed in the diff and by a test that asserts the body-reader mock is never invoked. The chunked-response residual is real (Playwright's `APIResponse` type has no streaming accessor — confirmed against the Playwright type defs) and is documented in-code exactly as claimed; this isn't fixable without a bigger rework, and the comment is honest about it being a correctness backstop, not a memory one.
- Read oracle (`presigned-upload.ts`): the store's response body is no longer returned to the caller on failure, only logged — confirmed in code and by test.
- `skip()` UTF-8 decode and `upload_headers` override: both fixed and covered by tests that check actual byte-level behavior (binary survives as base64, HTML stays utf-8; a caller-supplied `Content-Length`/`Content-Type` no longer wins over the computed ones while unrelated headers still pass through).
- Hash-in-pipeline refactor (`f58ca25`): correctly closes the theoretical 'data'-listener-vs-toWeb ordering gap, well covered by size-boundary tests.

**❓ Needs a human decision**
- See inline comment on `presigned-upload.ts:33` — I could not find `_validate_tabby_target_url` in tabby, adoptwebui, python-mcp, or adoptai-workflows. Is "mirrors the platform's own guard" accurate, or should that framing be corrected?
- See inline comment on `presigned-upload.ts:148` — the redirect bypass. Is blind SSRF to internal endpoints (no body echoed back) an acceptable residual risk, or should `redirect: 'manual'` be added now?

**🔒 Aikido scan:** Not run this pass — this was a targeted verification of the two prior blockers/concerns, not a full re-review.

apps/worker/src/presigned-upload.ts
siddhant-adoptai: Fixed in `cce6aa9` — this was a real bypass and the diagnosis was exact.

`fetch()` now passes `redirect: 'manual'` and any 3xx is thrown rather than followed. The guard validated where the *supplied* host resolved; undici re-issuing the PUT at a `Location` nothing had checked defeated it with a response header.

One thing worth recording: following a redirect here could never have worked anyway. The streaming path passes a `ReadableStream` body, which is not replayable, so a redirect on a `put_download` would have failed at the second hop regardless — the bypass was only ever reachable in a way that also broke the upload. That does not make it less of a hole (the request still reaches the internal address, which is the SSRF), but it means refusing costs nothing legitimate: no store needs this, and a 307 from a presigned URL means it was minted against the wrong endpoint. The error says that.

Test: `does not follow a redirect from the upload target` asserts both that `redirect: 'manual'` is passed and that a 307 is refused; a second unit test covers the stream being destroyed rather than leaked on that path. Verified non-vacuous — removing the option and the 3xx branch fails exactly those two tests and nothing else.

apps/worker/src/presigned-upload.ts
siddhant-adoptai: The function exists — `ProjectA3/actionbot/widdle_executor.py:2918`, `WiddleExecutor._validate_tabby_target_url`, tested at `tests/test_tabby_routed_rest.py:573`. ProjectA3 is the one repo not in the four that were grepped, which is a reasonable place to miss it since it is the caller rather than a peer service.

Worth confirming the claim is about substance and not just a name, so specifically what it mirrors: same scheme allowlist, same "check the literal host plus every address it resolves to" structure, same decision to let DNS-resolution failures through rather than fail closed, and for the same stated reason — the resolver here may differ from the one at the other end, and the NetworkPolicy is the hard boundary. It uses Python's `ipaddress` properties (`is_private`/`is_loopback`/`is_link_local`/`is_reserved`/`is_unspecified`/`is_multicast`) where this has to spell the ranges out, so the TS version enumerates CGNAT explicitly; that is the one place the two differ in coverage rather than expression.

No change made — flagging it as a question rather than assuming either way was the right call, and I would rather it get asked than taken on trust.

apps/worker/src/presigned-upload.ts
siddhant-adoptai: Both parts addressed in `cce6aa9`.

**Rebinding:** accepted, and now written down as accepted rather than left to look like an oversight. The docstring says what the gap is (resolve here, resolve again at connect), why it is not closed (Node's fetch does not let the connection be pinned to the address that was checked — that needs a custom agent or a resolved-IP-with-SNI rewrite, which is a different change), and what does not depend on resolver timing: the worker pod's NetworkPolicy. This guard is defence in depth in front of that, not a replacement for it.

**Test gap:** you were right that the resolution branch had no coverage — the added tests all passed literals, which exercise the `isIP(host)` path and never call DNS. New `presigned-upload.spec.ts` mocks `dns.lookup` and covers a public-looking hostname resolving to `169.254.169.254`; a round-robin record where only one of several answers is internal (the case a "does it resolve to something bad" check written as "first answer" would miss); an IPv6 unique-local answer; an all-public answer being accepted; and resolution failing being allowed through, so the fail-open choice is asserted rather than incidental.

Checked it discriminates: dropping just the `dnsLookup` block while leaving the literal check intact fails exactly the three refusal tests and leaves the rest green — so the new tests cover the branch that was untested, not the one that already worked.

navjot-adopt (changes requested):
Reviewed the full branch (all 7 commits) plus the API, controller proxy wiring and the helm chart. Ran the worker and API suites on a clean pnpm install of the branch: worker 360/360, API 681/681, worker typecheck clean. The OOM-ordering fix, the redirect refusal, the header-override fix and the skip() decode fix all check out against the code.

**Requesting changes for one blocker and one structural question.**

**Blocker: the upload-URL guard is bypassed by IPv4-mapped IPv6 literals.** Node normalises `[::ffff:127.0.0.1]` to the hex form `[::ffff:7f00:1]`, and the unwrap regex only matches the dotted form, so the mapped address is never re-checked. Probe against this branch's `validateUploadUrl`:

```
http://[::ffff:127.0.0.1]/   -> ALLOWED
http://[::ffff:7f00:1]/      -> ALLOWED
http://[::ffff:a9fe:a9fe]/   -> ALLOWED   (169.254.169.254, the metadata endpoint)
http://[64:ff9b::a9fe:a9fe]/ -> ALLOWED   (NAT64 well-known prefix)
http://169.254.169.254/      -> blocked
http://2130706433/           -> blocked
```

On a dual-stack pod a connect to an IPv4-mapped address reaches the IPv4 target. This is the same class as the redirect bypass the last round closed. Please stop hand-rolling the range check: `ipaddr.js` is already in the dependency tree (via express -> proxy-addr) and handles mapped, NAT64 and range classification. Note also that ProjectA3's `_validate_tabby_target_url` uses Python's `ipaddress`, which classifies `::ffff:a9fe:a9fe` as link-local correctly, so the two guards are not actually mirrors today.

**Structural question: the PUT bypasses the egress proxy by design.** The worker launches Chromium with `--proxy-server=EGRESS_PROXY_URL` (main.ts) and the egress proxy enforces a per-session host allowlist. The new PUT goes out through Node's global fetch, which does not use that proxy, so the allowlist does not apply to it at all. The PR leans on "NetworkPolicy stays the hard boundary", but there is no NetworkPolicy anywhere in helm-charts (grepped the whole repo). As deployed, the egress proxy IS the boundary, and this path steps around it. Two ways out: (a) send the PUT through the same proxy with undici's `ProxyAgent` (it streams a body fine, so the buffering objection to `context.request` does not apply) and add the bucket host to the allowlist, or (b) link the NetworkPolicy from whichever infra repo it lives in so the claim is checkable. I would prefer (a).

**Minor, not blocking**
- `skip()` still calls `resp.body()` on a non-2xx with no size check, so an oversized error body has the same unbounded-read shape as the chunked case. Worth a content-length pre-check there too.
- The 2560Mi worker figure in the MAX_SINK_BODY_BYTES comment does not match the chart: adopt-tabby values set 1Gi limits and I found no worker override. Either the comment or the chart is stale.
- `download_url` on a non-attachment URL waits the full timeout before reporting, since the download event never fires. Acceptable, but a caller will see it as a hang.

CI green on all 7 checks. #176 compatibility already verified by the bot against current dev; I did not redo it.

siddhant-adoptai:
Both blockers fixed in `55ba79b`. Worker 373/373, API 681/681, typecheck and lint clean, `pnpm install --frozen-lockfile` verified.

**The IPv4-mapped bypass — confirmed exactly as described.** Reproduced before touching anything: `new URL('http://[::ffff:127.0.0.1]/').hostname` is `::ffff:7f00:1`, and the unwrap regex only matched the dotted form, so it never fired.

Taking the `ipaddr.js` suggestion, and going further than swapping the parser: the rule is now an **allowlist** — anything not ordinary public `unicast` is refused. That covers every transitional encoding that embeds an IPv4 address (`ipv4Mapped`, `rfc6052`/NAT64, `rfc6145`, `6to4`, `teredo`) without needing a case for each. Worth saying why rather than just that it is shorter: a blocklist of known-bad ranges is the shape that let *both* this and the redirect bypass through, and enumerating the encodings I currently know about would have been the same shape a third time.

One correction to the premise, because it would have bitten whoever did this next: **`ipaddr.js` was not usable from the worker as-is.** It is in the tree only transitively (via `minio`, and `proxy-addr` pulls a separate 1.x), and pnpm's layout does not make a transitive dependency resolvable — `require.resolve('ipaddr.js')` from `apps/worker` fails. It is now declared explicitly at `^2.3.0`, which is the version `minio` already pins, so nothing new is installed.

Tests: the five probe vectors from your review as URL literals, so they go through the same `URL.hostname` normalisation that defeated the old check, plus a hostname that *resolves* to a mapped address, plus a public IPv6 literal so the rule is not "refuse all IPv6". Verified non-vacuous — against the previous implementation exactly those six fail and the seven pre-existing ones pass.

**The egress proxy — you are right, and the claim the PR leaned on was false.** I grepped for it before arguing: `charts/browser-hitl/templates/network-policies.yaml` defines policies for api, controller, postgres, redis and nats, and **nothing selects the worker component**. So there was no NetworkPolicy to link, the egress proxy is the boundary as deployed, and this PUT stepped around it.

Went with (a). The PUT now goes out on an undici `ProxyAgent` pointed at `EGRESS_PROXY_URL` — the same variable `main.ts` launches Chromium behind — with credentials moved into the CONNECT's `Proxy-Authorization`, which is how the proxy attributes a session. One agent is cached per proxy URL rather than built per upload. Unset `EGRESS_PROXY_URL` (local dev) still goes direct, and an *unparseable* one now throws instead of silently going direct, since failing open there would quietly reopen the hole this closes.

Two things worth knowing before this ships:

- **The bucket host has to be in the session allowlist.** A presigned URL is not self-authorising as far as egress is concerned, and `TABBY_INFRASTRUCTURE_ALLOWLIST` only covers `.adopt.ai`. A misconfigured allowlist will now surface as a proxy refusal rather than a store error. I have documented it at the call site; whether the S3 host goes in the default allowlist or is pushed per-session is your call, and it is an ops change either way.
- `undici` is a **new direct dependency**. Node's builtin fetch is undici underneath, but `ProxyAgent` is not reachable from any builtin, so there was no zero-dependency route to this.

The stale rebinding comment naming the NetworkPolicy is corrected rather than left to be believed — the proxy decides by hostname, so unlike a resolved-address check it is not defeated by a second resolution.

**On the three minor points**

- `skip()` now refuses on the declared `content-length` before reading, same ordering rule as the success path. Found something while doing it: `overSink` was declared *after* `skip` but called from inside it, and `skip` runs at two sites above that declaration — so the throw would have been a `ReferenceError` from the temporal dead zone rather than a 413. Declaration hoisted above `skip`. Two tests: one asserts the body function is never invoked, one asserts a normally-sized diagnostic still comes back.
- **The 2560Mi figure — partly.** The chart default *is* 2560Mi (`values.yaml:130`); 1Gi is the memory **request**, not the limit. But the deployed value is neither: `infra/tfy/deploy.yaml` takes `${WORKER_MEM_LIMIT}`, which both deploy workflows default to **3Gi**. Since it is deployment-set, the comment no longer names a number and says the ceiling is deliberately under the smallest of them.
- `download_url` no longer waits the full timeout. A navigation that *rendered* is the not-an-attachment case — Chromium aborts the navigation when a response goes to the download manager — so after a clean load it waits a 5s grace for a script-started download rather than the command timeout, and says which case it hit. The grace is there because a page starting a download from script after load is real; waiting minutes for it is not.

Not changed: the chunked-response residual, which is still Playwright's `APIResponse` having no streaming accessor.

navjot-adopt (dismissed):

rahulbh1510 (approved):

PR #182, opened 2026-09-22
feat(execute): allow upload_url on /execute/fetch for presigned download sink

siddhant-adoptai:
## What

`/execute/fetch` can now stream a downloaded file straight to a presigned PUT URL, the same way `/execute/browser` `put_download` already can.

## Why

The worker, the `ExecuteService`, and the shared `ExecuteFetchRequest` type already fully support `upload_url` / `upload_headers` / `upload_always` — the worker validates the sink URL and PUTs the bytes, the service forwards and validates them. But `ExecuteFetchDto` never declared these fields. Because the API mounts a global `ValidationPipe({ whitelist: true, forbidNonWhitelisted: true })`, any caller sending `upload_url` was rejected with a 400 before the handler ran, so the S3 download sink was unreachable through `/execute/fetch`.

This is the follow-up gap noted in the #178 review: the sink plumbing merged, but the fetch DTO was never opened up, so only the browser path could use it. It blocks the Workday report-file download from returning an S3 link instead of a session-bound Workday URL.

## Change

- Declare `upload_url` (`@IsString`), `upload_headers` (`@IsObject`), `upload_always` (`@IsBoolean`) on `ExecuteFetchDto`, all optional.
- Forward the three fields from the DTO into `executeService.executeFetch(...)`.
- `export` the DTO and add `execute-dto.spec.ts`, which drives the real `ValidationPipe` (same config as `main.ts`) to prove the sink payload passes, a plain fetch still works, an unknown field is still rejected, and a non-string `upload_url` is rejected.

No behaviour change for callers that do not send the new fields.

## Testing

- `execute-dto.spec.ts` + `execute.service.spec.ts`: 21 passed
- `tsc --noEmit` clean; pre-commit ran `nx run-many --target=build --all` (7 projects) — all green


rahulbh1510 (commented):
**Verdict:** No blockers or concerns found — small, well-scoped, additive-only change with matching test coverage at the layers this repo actually tests (DTO/ValidationPipe + service). Safe to merge pending CI.

**🔒 Aikido scan:** No open findings for this PR (scan matches head commit de3a616).
**✅ CI/Checks:** Could not be checked from this token (Checks API is unavailable to fine-grained PATs) — confirm green on the PR's Checks tab before merge.

navjot-adopt (approved):

rahulbh1510 (approved):

PR #183, opened 2026-09-22
fix(execute): accept the sink timeout ceiling on ExecuteFetchDto

siddhant-adoptai:
## What

`ExecuteFetchDto.timeout_ms` is capped at `EXECUTE_LIMITS.MAX_SINK_TIMEOUT_MS` (300s) instead of `MAX_TIMEOUT_MS` (60s).

## Why

#182 declared `upload_url` on the DTO so the presigned download sink could be reached through `/execute/fetch`. It could not: a fetch-to-sink caller sends the sink budget, because the worker is streaming a whole export into the object store rather than making an API call. The DTO still capped `timeout_ms` at the API ceiling, so the global `ValidationPipe` rejected **every** sink request with a 400 before the handler ran.

That is the 400 the Adopt executor has been getting on stage. It also explains the contrast that made it look like a bad `upload_url`: a hand probe of the same Workday URL with `timeout_ms` under 60s passed validation and reached the PUT (502 `upload of 11,632 bytes failed`), while the real WDL step — which asks for the full 300s — never got past the pipe.

Both ceilings already exist and are already applied correctly one layer down. `execute.service.ts` clamps a plain fetch to `MAX_TIMEOUT_MS` and a sink fetch to `MAX_SINK_TIMEOUT_MS` (covered by `execute.service.spec.ts`), so raising the DTO bound widens nothing for an ordinary API call — a plain fetch asking for 300s is still clamped to 60s by the service.

## Tests

`execute-dto.spec.ts` gains three cases: a sink payload asking for the sink ceiling passes the pipe; one above it is still rejected; a plain fetch with a large timeout passes the pipe and is left to the service clamp.

`make test` green across api (688), worker (419) and controller (90). `make lint` clean.


rahulbh1510 (commented):
**Verdict:** Don't block — the core fix is correct and well-tested; the two Concerns are worth a follow-up but neither breaks anything as shipped.

**❓ Needs a human decision**
- The PR description's diagnosis of `TabbyFetchClient._request` (raise_for_status dropping the validation error body, treating 400 as retryable) doesn't correspond to any file in this repo — no `TabbyFetchClient` or `raise_for_status` call exists here. That component appears to live in a different codebase (the calling executor). Is that fix tracked separately, and should this PR's description be scoped to not imply it's fixed here?

**🔒 Aikido scan:** No open findings for this PR (scan matched the reviewed commit).
**✅ CI/Checks:** Could not be checked from this token; confirm on the PR's Checks tab

apps/api/src/modules/execute/execute.controller.ts
```diff
@@ -35,7 +35,11 @@ export class ExecuteFetchDto {
   @IsOptional()
   @IsInt()
   @Min(1000)
-  @Max(EXECUTE_LIMITS.MAX_TIMEOUT_MS)
+  // The sink ceiling, not the API one: a fetch-to-sink caller asks for up to
+  // MAX_SINK_TIMEOUT_MS and the service clamps a plain fetch back to MAX_TIMEOUT_MS.
+  // Capping here at the smaller value rejected every sink request with a 400
+  // before the handler ran.
+  @Max(EXECUTE_LIMITS.MAX_SINK_TIMEOUT_MS)
```
rahulbh1510: **⚠️ Concern**

The widening to `MAX_SINK_TIMEOUT_MS` is unconditional, not gated on `upload_url` being set. A plain (non-sink) fetch asking for e.g. 200s used to get an explicit 400 at the DTO layer; now it passes validation and is silently clamped by `execute.service.ts` to 60s, so the caller finds out via an eventual timeout instead of an immediate, actionable 400. `@ValidateIf(o => !o.upload_url)` (keeping the 60s cap for non-sink requests) would preserve the old fail-fast behavior for ordinary callers while still unblocking sink requests.

apps/api/src/modules/execute/execute.controller.ts
```diff
@@ -97,7 +101,10 @@ class ExecuteBrowserDto {
   @IsOptional()
   @IsInt()
   @Min(1000)
-  @Max(EXECUTE_LIMITS.MAX_TIMEOUT_MS)
+  // Same reason as ExecuteFetchDto: put_download streams a captured file to an
+  // object store and the service gives that command MAX_SINK_TIMEOUT_MS, which a
+  // caller could not ask for while this capped at the API ceiling.
+  @Max(EXECUTE_LIMITS.MAX_SINK_TIMEOUT_MS)
```
rahulbh1510: **⚠️ Concern**

Same unconditional-widening issue as `ExecuteFetchDto.timeout_ms` above: any browser command (not just `put_download`) can now request up to 300s at the DTO layer and be silently clamped to 60s by the service instead of getting an explicit 400. Same fix would apply here (`@ValidateIf(o => o.command !== 'put_download')` or similar).

apps/api/src/modules/execute/execute-dto.spec.ts
```diff
@@ -50,4 +51,71 @@ describe('ExecuteFetchDto — presigned upload fields', () => {
...
+    );
+    expect(out.upload_url).toBeUndefined();
+    expect(out.timeout_ms).toBe(EXECUTE_LIMITS.MAX_SINK_TIMEOUT_MS);
+  });
+});
+
+// put_download is the browser route's sink: execute.service gives that command
+// MAX_SINK_TIMEOUT_MS, so the DTO has to let a caller ask for it — the same
+// 400-before-the-handler this PR fixes on /execute/fetch.
+describe('ExecuteBrowserDto — sink timeout ceiling', () => {
```
rahulbh1510, line 106: **💡 Suggestion**

The `ExecuteFetchDto` block gets a third test proving a plain (non-sink) call still passes the pipe and is left to the service clamp. The `ExecuteBrowserDto` block only has the "accepts sink ceiling" and "rejects above ceiling" cases — no equivalent "a plain (non-`put_download`) command with a large timeout still passes the DTO pipe" case, so the asymmetry with the widening behavior above isn't pinned at this layer for the browser DTO (only covered at the service layer in `execute.service.spec.ts`).

apps/api/src/modules/execute/execute.controller.ts
siddhant-adoptai: Fixed in `607c364`. Agreed this was a real regression, and on reflection it is the conservative fix rather than a DX nicety: before this branch a plain fetch asking 200s got a 400, and the blanket widening quietly turned that into a 60s clamp the caller only learns about as a timeout.

`@Max` is static and `@ValidateIf` gates every validator on the property, so neither expresses a per-mode bound alone. Added a small local `MaxTimeoutForMode` constraint that picks the ceiling from the request itself — `upload_url` present means the sink ceiling, otherwise the API one. The change is now strictly additive: every request valid before this branch is still valid, with the same error when it is not. New test pins that a plain fetch asking for the sink ceiling is still rejected.

apps/api/src/modules/execute/execute.controller.ts
siddhant-adoptai: Fixed in `607c364` by the same constraint, gated on `o.command === 'put_download'` — only that one command gets the larger budget and every other command keeps the API ceiling.

apps/api/src/modules/execute/execute-dto.spec.ts
siddhant-adoptai: Added in `607c364`. The browser block now has the symmetry case — a `navigate` asking for the sink ceiling is rejected — plus one showing it accepted at the API ceiling. Note the assertion flipped from the fetch block's original: with the per-mode ceiling these are refusals rather than pass-through-and-clamp.

siddhant-adoptai:
On the open question from the review: the `raise_for_status()`-drops-the-body problem is **not** in this repo — correct that grepping tabby finds nothing. It lives in the caller, `ProjectA3/actionbot/operations/tabby_fetch.py::TabbyFetchClient._request`, which treated a 400 as transient, retried it four times, and then built its error with `raise_for_status()` — discarding the body where Tabby names the field it rejected. That is why this took days to root-cause: the answer was in every response and thrown away each time.

It is **fixed and tracked**, in the paired PR: adoptai/ProjectA3#1566. That PR maps every 4xx to an error carrying `resp.text`, stops retrying refusals, and keeps retrying 408/429 (which say "not now", not "not this") with their reason preserved into the give-up message. So it is not an open bug anywhere — just one repo over.

rahulbh1510 (commented):
**Verdict:** All three prior findings are resolved (two fixed in code, one confirmed moot/tracked elsewhere) and no new blocking issues found — safe to merge once CI is confirmed green.

**Prior findings status:**
1. ⚠️ Unconditional `MAX_SINK_TIMEOUT_MS` widening — **Addressed.** New `MaxTimeoutForMode` custom validator (execute.controller.ts) gates the ceiling per-mode: `upload_url` truthy for `ExecuteFetchDto`, `command === 'put_download'` for `ExecuteBrowserDto`. This matches execute.service.ts's own conditions exactly (lines computing `timeoutMs` via `request.upload_url ? MAX_SINK_TIMEOUT_MS : MAX_TIMEOUT_MS` and `request.command === 'put_download' ? ... : ...`). A plain fetch/non-put_download command asking for >60s now gets an explicit 400 again, confirmed by new tests.
2. 💡 Missing symmetric test for `ExecuteBrowserDto` — **Addressed.** Two new tests added ("rejects a non-put_download command asking for the sink ceiling", "accepts a non-put_download command at the API ceiling") mirroring the fetch DTO's block.
3. ❓ `TabbyFetchClient._request`/`raise_for_status()` question — **Moot for this repo, now confirmed.** Author's PR comment confirms the bug is in `ProjectA3/actionbot/operations/tabby_fetch.py::TabbyFetchClient._request` (retried 4xx as transient, then dropped the response body via `raise_for_status()`), fixed and tracked in adoptai/ProjectA3#1566 (verified: open PR, correct title/scope). Confirmed tabby repo still has no trace of this code, as expected.

**Diff scope:** Only 2 files changed per PR metadata (`execute.controller.ts`, `execute-dto.spec.ts`), matching the stated intent. All 3 commits are `fix(execute)`-prefixed and on-topic; base is `dev` with no unrelated commits.

**🔒 Aikido scan:** No open findings for this PR (empty findings array).

navjot-adopt (approved):

rahulbh1510 (approved):

PR #184, opened 2026-09-23
fix(worker): let fetch derive Content-Length for a Buffer upload

siddhant-adoptai:
## Problem

Every `/execute/fetch` sink upload failed before the request left the worker pod:

```
/execute/fetch: upload of 571 bytes failed: fetch failed (invalid content-length header)
```

`uploadToPresignedUrl` set `'Content-Length': String(sizeBytes)` **and** handed `fetch` a Buffer
body, which fetch derives the length from itself. The header list then held `"571, 571"`; undici
validates content-length as all-digits, so the comma made it an *invalid* content-length (not a
duplicate) and the PUT was never dispatched.

Reproduces on Node 20 and 22 whenever a dispatcher is in play — which `55ba79b` (send the PUT
through the egress proxy) made unconditional. It was masked until now by the DTO timeout cap
(#183), which rejected these requests before the worker ran at all.

## Fix

Send `Content-Length` only for the stream body. A Buffer body still arrives with the correct
length (verified against a real dispatcher); a stream has no derivable length and a presigned PUT
is signed for one, so that path must keep sending it. A caller-supplied value is still dropped and
still cannot reach the store.

## Why the existing tests missed it

Every test in `presigned-upload.spec.ts` replaces `global.fetch` and asserts on `init.headers` —
the one place this looks correct. The new test stands up an origin plus a CONNECT proxy, dispatches
for real, and asserts what the origin receives. It fails without the fix.


rahulbh1510 (commented):
**Verdict:** No blockers or concerns found — clean, well-scoped fix with a real integration test that reproduces the original bug. Fine to merge as-is.

**🔒 Aikido scan:** No open findings for this PR (tabby).

navjot-adopt (dismissed):

navjot-adopt (approved):

rahulbh1510 (approved):

